PR #277, opened 2021-02-03
Proposal: New header/nav design

sofiarm21:
### **Issue #266* - Proposal*

**Description:**

This pr contains a new design proposal for the `navigation` component, instead of using `header` we only render `navigation` with the name of the current page or project.

**Breakdown:**

* A new reactive variable `pageName` that stores the name of the current page
* The `navigation` component reads the value of  `pageName` reactive variable and displays it as title, if the value of `pageName` is null it uses the location url pathname

**Demonstration:**

Screen Shot 2021-02-03 at 12 28 35 PM
Screen Shot 2021-02-03 at 12 28 16 PM
Screen Shot 2021-02-03 at 12 27 50 PM
Screen Shot 2021-02-03 at 12 27 35 PM

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -31,7 +31,7 @@ class App extends React.Component {
                 {
                     IS_PRODUCTION
                         ? <Navigation/>
-                        : <SwipeableNavigation/>
+                        : <Navigation pageName={''}/>
                 }
                 <ThemeProvider theme={theme}>
                     <PrivateRoute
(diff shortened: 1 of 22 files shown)


sofiarm21:
@otech47 you like this better?

Screen Shot 2021-02-17 at 12 46 17 PM

otech47:
@sofiarm21 new conflicts here

Issue #279, opened 2021-02-04
Using async/await inside useEffect() causes random bug


sofiarm21:
<img width="1440" alt="Screen Shot 2021-02-04 at 1 53 39 PM" src="https://user-images.githubusercontent.com/49292858/106934527-98ff8980-66f0-11eb-900d-2dc2e84757f3.png">

PR #280, opened 2021-02-04
Bug resolved: project overview async

sofiarm21:
### **Issue #279* - Bug resolved*

**Important:**

By mistake I named the branch referencing issue #270 but this resolves issue #279 

**Description:**

This pr contains the necessary changes to fix the bug described in #279

**What was happening:**

When going from `ProjectContributors` to another page a bug happened breaking the app, capture of the bug is in the issue #270 description. This was happening because `async/await` shouldn't be used in the `useEffect` function like what I was doing, and this was causing the bug.

**How I fixed it:**

I reimplemented the code to keep it functional deleting the `async/await` and using another `useEffect` function

**Magnitud:**

The bug was on `develop` so probably the deploy has it too, this could be breaking the app for the testing users even that no one has reported it yet

**Proof of resolved bug:**

Before: https://www.loom.com/share/44375b6410034f7baef85f6768b76454
After: https://www.loom.com/share/90656d184d0a4447bddacd2bcb6d0450

### DIFF
--- a/src/components/ContributorsEmptyState.js
+++ b/src/components/ContributorsEmptyState.js
@@ -5,7 +5,7 @@ import {
     Icon,
     Typography
 } from '@material-ui/core'
-import SupervisedUserCircleIcon from '@material-ui/icons/SupervisedUserCircle';
+import SupervisedUserCircleIcon from '@material-ui/icons/SupervisedUserCircle'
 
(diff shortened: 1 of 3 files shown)


otech47 (approved):
https://www.loom.com/share/7144153def8544bc8152636be22371c2

Issue #285, opened 2021-02-06
Amount in AddAllocation is not correct 


sofiarm21:
<img width="1348" alt="Screen Shot 2021-02-06 at 12 02 24 PM" src="https://user-images.githubusercontent.com/49292858/107123884-9d55af00-6876-11eb-974c-c5ac90ca963a.png">

Issue #263, opened 2021-01-27
Tweak the active contributors list


otech47:
- Show the contributor if they have at least 1 allocation (even if not active or proposed/active)
- Clicking the contributor tile should do the same as if they were in the list below

PR #287, opened 2021-02-06
Active contributors list

sofiarm21:
### **Issues #263 #285**

**Important:**

I merged branch from pr #282 into this branch

**Description:**

This pr contains the necessary changes for resolving issues #263 and #285. 

Rewarding issue #263, the contributors that have active and upcoming allocations get rendered on the list at the top of the page, divided by active and upcoming allocations, now it's possible to add an allocative for a contributor that has an a active allocation ant the contributor tile is expandable to see the details of it's allocations

**Breakdown:**

* I created a new util function `formatAmount` that basically makes the formatting using `accounting.js` but I transformed it into it's own function for better understanding and less repetitive code
* New util function `getActiveAndUpcomingAllocations ` does basically what the function name says and optionally returns only the active or the upcoming allocations from a given set of allocations. I made this a function because the logic was used multiple times with different parameters so I decided to combine everything into one reusable function.
* The top list of the `ProjectsContributors` page now renders contributors that has active and upcoming allocations, divided by active and upcoming contributors, If a contributor has an active allocation and a upcoming allocation it will be rendered on the active contributors sections. The function `upcomingContributorsAllocatedOnly` is the one that implements this logic, basically takes the contributors with upcoming allocations and filters so we only render those with upcoming allocations.
*  `GET_ALLOCATIONS` query gets the allocations passing the project_id and contributor_id parameters, this data is used later on the `contributorTile` to render in an accordion the allocations of each contributor with it's information
* Now the `ContributorTile` is expandable, showing the allocation that the contributors has for the particular project, it only shows active and upcoming allocations so it doesn't render the past ones, and orders them by date so the active one is going to be the first one of the list and will be highlighted . The `isActiveAllocation ` constant on the `ContributorTIie` is for dynamically apply certain css properties if the allocation is the active one
* I also made general cleanup and little bug fixes

**Proof of implementation:**

* Issues #263 

https://www.loom.com/share/a7f9261829574e90807fdbe7764b70c9

* Issue #285

Screen Shot 2021-02-06 at 2 05 18 PM

### DIFF
--- a/src/scripts/selectors.js
+++ b/src/scripts/selectors.js
@@ -1,22 +1,8 @@
-import { find } from 'lodash'
+import { filter, find } from 'lodash'
 import accounting from 'accounting-js'
+import moment from 'moment'
 
 import { CURRENCIES } from '../constants'
-
-export const selectCurrencySymbol = (props) => {
-    const {
-        currency
-    } = props
-    return find(CURRENCIES, c => {
-        return c.name == currency
-    })['symbol']
-}
-
-export const selectCurrencyInformation = (props) => {
-    return find(CURRENCIES, c => {
-        return c.name == props.currency
-    })
-}
 
 export const formatAmount = (props) => {
     const {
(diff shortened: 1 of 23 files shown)


src/scripts/selectors.js
```diff
@@ -1,10 +1,67 @@
...
+            }
+        } else if (upcomingOnly) {
+            if (moment(a['start_date'], 'x').isAfter(moment())) {
+                desiredAllocations.push(a)
+            }
+        } else if (moment(a['end_date'], 'x').isAfter(moment())) {
+            desiredAllocations.push(a)
+        }
+    })
+    return desiredAllocations
```
otech47, line 52: Just FYI consider using `allocations.filter` for this if there are not transformations needed on the objects in the array.

(Or `reduce` if both filtering and re-mapping/transforming is needed)

otech47 (approved):

Issue #290, opened 2021-02-09
Build Contributor Detail Page


otech47:
- Add a button on ContributorTiles to be able to navigate to this new route `/contributor/:contributorId`
- Include a form to be able to add a Toggl ID to the Contributor

PR #298, opened 2021-02-10
Contributor Detail Page

sofiarm21:
### **Issue #290**

**Important:**

For implementing this I merged pr #287 and #297, I recommend reviewing this first

**Description:**

This pr contains the necessary changes for implementing the contributor detail page. The contributor Detail page has the contributor information, including the project where the contributor has been involved and the allocations that have received

**Breakdown**

A lot is happening in this pr
* The route `/contributor/:contributorId` is the one that renders the `ContributorDetailPage` and for now is only accessible through the `ProjectContributorsPage` or directly from the URL
* We can propose an allocation for the selected contributor, for this I had to make changes on the `AllocationAddForm` component and I created a new component called `AllocationProposeSpecifics` that makes the same as `AllocationAddSpecifics` with the difference that doesn't receive a project as a parameter but a contributor instead and fetches all project and payment dynamically.
* On the Propose allocation form, compared to the allocation form from the ProjectContributor/ProjectPayemtns page is different because it doesn't render the entire form at once, this is because is waiting for the user to select a client through a project to know exactly which currency should display
* The logic 
```
`${!allocation.date_paid
                                ? `${grey}`
                                : futureAllocation
                                    ? `${orange}`
                                    : `${setlifeBlue}`
                            }`
```
is implemented to conditional render the start date and end date in each AllocationTile in different colors, as detailed on the figma file [https://www.figma.com/file/qgGWXmprU7vTv7guzWzvML/Project-Trinary?node-id=1716%3A756](url)

**Important to notice**

* Rn the total paid is the sum of all the payments in the USD currency, I opened a new issue that will group the total paid by the currency type
* The tile of project involvement is the same `ProjectTile` component reused in other pages, we are not making any UI difference to now if it's actively involved in the project or if he worked in the past, would be nice to work on this on a future implementation
 
**Proof of implementation**

https://www.loom.com/share/b219757890cc41199cb22089e9ad65c0

### DIFF
--- a/src/components/AllocationTile.js
+++ b/src/components/AllocationTile.js
@@ -66,11 +66,11 @@ const AllocationTile = (props) => {
                 <Box my={5}/>
                 <Grid item xs={6}>
                     <Box color={
-                        `${allocation.date_paid
-                            ? 'primary'
+                        `${!allocation.date_paid
+                            ? 'seccondary.light'
                             : futureAllocation
-                                ? 'yellow'
-                                : 'secondary'
+                                ? `${orange}`
+                                : 'primary.main'
                         }`}
                     >
                         <Typography
--- a/src/components/AllocationTile.js
+++ b/src/components/AllocationTile.js
@@ -1,0 +1,115 @@
+import React from 'react'
+import {
+    Box,
+    Card,
+    Grid,
+    Typography
+} from '@material-ui/core'
+import AccountBalanceWalletIcon from '@material-ui/icons/AccountBalanceWallet'
+import accounting from 'accounting-js'
+import moment from 'moment'
+
+import { selectCurrencyInformation } from '../scripts/selectors'
+import { orange, red } from '../styles/colors.scss'
+
+const AllocationTile = (props) => {
+
+    const {
+        allocation
+    } = props
+
+    const currencyInformation = selectCurrencyInformation({
+        currency: 'USD'
+    })
+    //TODO: Replacethe following function with upcoming selecto formatAmount (currently on different pr)
+    const formatPaymentAmount = (props) => {
+        const { amount, currencyInformation } = props
+        return accounting.formatMoney(
+            amount,
+            {
+                symbol: currencyInformation['symbol'],
+                thousand: currencyInformation['thousand'],
+                decimal: currencyInformation['decimal'],
+                format: '%s %v'
+            }
+        )
+    }
+    const paymentAmount = formatPaymentAmount({
+        amount: allocation.amount / 100,
+        currencyInformation: currencyInformation
+    })
+    const futureAllocation = moment(allocation.start_date, 'x').isAfter(moment(), 'days')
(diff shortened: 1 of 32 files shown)


src/components/AllocationTile.js
```diff
@@ -0,0 +1,109 @@
...
+
+const AllocationTile = (props) => {
+
+    const {
+        allocation
+    } = props
+
+    const currencyInformation = selectCurrencyInformation({
+        currency: 'USD'
+    })
```
otech47, line 26: Suggestion, I prefer this pattern:

```
const AllocationTile = ({
    allocation
}) => {
    const currencyInformation = selectCurrencyInformation({
        currency: 'USD'
    })

    ...
```

src/components/AllocationTile.js
```diff
@@ -0,0 +1,109 @@
...
+    const paymentAmount = formatAmount({
+        amount: allocation.amount / 100,
+        currencyInformation: currencyInformation
+    })
+    const futureAllocation = moment(allocation.start_date, 'x').isAfter(moment(), 'days')
+    const weeksOfdDifference = moment(allocation.end_date, 'x').diff(moment(), 'weeks')
```
otech47, line 32: Typo here `weeksOfdDifference`

src/components/ContributorProjectsCollab.js
```diff
@@ -0,0 +1,71 @@
...
+    Card,
+    Grid,
+    Typography
+} from '@material-ui/core'
+
+import LoadingProgress from './LoadingProgress'
+import ProjectTile from './ProjectTile'
+import { GET_CONTRIBUTOR_PROJECTS } from '../operations/queries/ContributorQueries'
+
+const ContributorProjectsCollab = (props) => {
```
otech47, line 14: Not sure about this component name (remember to [avoid abbreviations](https://github.com/setlife-network/trinary/blob/develop/CONTRIBUTING.md#3-avoid-abbreviations))

Suggested change to `ContributorProjectInvolvement`? Or just `ContributorInvolvement`?

src/components/ContributorTile.js
```diff
@@ -1,53 +1,241 @@
...
+        contributor,
+        onAddButton,
+        project
+    } = props
+    const history = useHistory()
 
     const handleAddButton = () => {
         onAddButton({ contributor })
     }
+    const redirectToContributorDetails = ({ contributor }) => {
```
otech47, line 38: Suggestion: change to `goToContributorDetails`

src/components/ContributorTile.js
```diff
@@ -1,53 +1,241 @@
...
+            contributorId: contributor.id
+        }
+    })
+
+    const renderContributorAllocations = (props) => {
+        const {
+            allocations
+        } = props
+
+        const currenctAndUpcomingAllocations = filter(allocations, (allocation) => (
```
otech47, line 58: Type here: `currenctAndUpcomingAllocations`

otech47 (approved):

PR #299, opened 2021-02-10
Contributor total paid and UI fixes

sofiarm21:
### **Issue #297**

**Description:**

This pr contains the necessary changes to be able to get from the DB the sum of the amount of the allocations made to a contributor. The reason why I implemented this is because this amount has to be displayed on the Contributor detail page. So will be merged into pr #298.

**Breakdown:**

* the calculated attribute `total_paid` was added into the `Contributor` type

**Proof of implementation:**

Screen Shot 2021-02-10 at 12 31 31 AM

### DIFF
--- a/api/schema/types/ContributorType.js
+++ b/api/schema/types/ContributorType.js
@@ -5,6 +5,7 @@ module.exports = gql`
     type Contributor {
         id: Int!
         toggl_id: Int
+        total_paid: Int!
         name: String!
         external_data_url: String
         github_id: String!
--- a/api/schema/resolvers/ContributorResolver.js
+++ b/api/schema/resolvers/ContributorResolver.js
@@ -17,6 +17,15 @@ module.exports = {
                 }
             })
         },
+        total_paid: async (contributor, args, { models }) => {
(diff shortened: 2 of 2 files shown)


otech47 (approved):

api/schema/types/ContributorType.js
```diff
@@ -5,6 +5,7 @@ module.exports = gql`
     type Contributor {
         id: Int!
         toggl_id: Int
+        total_paid: Int!
```
otech47, line 8: One pattern here I would recommend that I like to do is separate properties that correspond to a column in the DB from the custom computed properties like this one or even from relationships. This way it is not confusing just by looking at the Contributor type I might think that `total_paid` is an actual database column that would need to be constantly updated.

Example:

```
type Contributor {
    // Columns
    
    id: Int!
    toggl_id: Int
    name: String!
    external_data_url: String!
    ...
    
    // Computed properties/relations
    
    total_paid: Int!
    allocations: [Allocation]
    permissions: [Permission]
    timeEntries: [TimeEntry]
    rates: [Rate]
}

```

otech47 (changes requested):

Issue #275, opened 2021-02-03
Add timeframe option for Expected Budget


otech47:
Would be nice if for Projects you could not only set a number for `expected_budget` but also an option for `daily`, `weekly`, `quarterly`, `monthly`, or `yearly` that maybe would be in a dropdown right next to the box in the form

PR #310, opened 2021-02-13
Expected budget timeframe

sofiarm21:
### **Issue #275**

**Important:**

I merged pr #287 because I needed a function implemented in that branch, I recommend reviewing that first

**Description:**

This pr contains the necessary changes for adding an `expected_budget_timeframe` into each project, an all the process that involves the project, as creating editing, and DB storage.

**Breakdown:**

* I created a migration named `addExpectedBudgetTimeframeToProjects` to add the attribute `expected_budget_timeframe` into the DB. To make this I had to create a new config file and a `.sequelizerc` file for config the migrations
* I added the necessary changes in the `ProjectType` and `ProjectResolver` to be able to fetch and mutate this new attribute
* I added a selector to choose between different options as a timeframe when creating/ editing a project. These are `daily `, `weekly`, `monthly`, `yearly`
* The timeframe is showed in the `ProjectInfo` component next to the budget amount

**Testing:**

I tested the following scenarios

- [x] Create a project adding a timeframe and store info correctly into the DB
- [x] Edit a project and change the timeframe and update info correctly into the DB
- [x] Leave the timeframe empty when creating/editing a project, and not being able to process with the DB transaction

**Proof of implementation:**

https://www.loom.com/share/688a0cc23bc1422f8b65f55da148a5c8

### DIFF
--- a/api/config/config.js
+++ b/api/config/config.js
@@ -8,11 +8,11 @@ const {
 } = MYSQL
 
 module.exports = {
-    'development': {
-        'username': `${DB_USERNAME}`,
-        'password': `${DB_PASSWORD}`,
-        database: `${DB_NAME}`,
-        host: `${DB_HOST}`,
+    development: {
+        username: DB_USERNAME,
+        password: DB_PASSWORD,
+        database: DB_NAME,
+        host: DB_HOST,
         dialect: 'mysql'
     }
 }
--- a/api/config/config.js
+++ b/api/config/config.js
@@ -1,0 +1,18 @@
+const { MYSQL } = require('./credentials');
+
+const {
+    DB_HOST,
+    DB_USERNAME,
+    DB_PASSWORD,
+    DB_NAME
+} = MYSQL
+
+module.exports = {
+    'development': {
+        'username': `${DB_USERNAME}`,
+        'password': `${DB_PASSWORD}`,
+        database: `${DB_NAME}`,
+        host: `${DB_HOST}`,
+        dialect: 'mysql'
+    }
+}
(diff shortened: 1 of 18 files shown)


api/config/config.js
```diff
@@ -0,0 +1,18 @@
...
+    DB_HOST,
+    DB_USERNAME,
+    DB_PASSWORD,
+    DB_NAME
+} = MYSQL
+
+module.exports = {
+    'development': {
+        'username': `${DB_USERNAME}`,
+        'password': `${DB_PASSWORD}`,
```
otech47, line 13: I suggest enforcing uniform properties here for this object so that it's not inconsistent some properties having quotes and some don't. The only time quotes are ever really needed is if the property key has a space or hyphen or other special character

api/migrations/20210212200719-addExpectedBudgetTimeframeToProjects.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
```
otech47, line 1: Maybe we should add some documentation to the README as well for how to generate these migrations using sequelize. Even if it is just a link to the Sequelize documentation but some guidance would be helpful

src/constants/index.js
```diff
@@ -23,6 +23,29 @@ export const CURRENCIES = [
...
         thousand: ','
+    },
+    {
+        name: '',
+        symbol: '',
+        decimal: '.',
+        thousand: ','
+    }
+]
+export const EXPECTED_BUDGET_TIMEFRAME_OPTIONS = [
```
otech47, line 34: Maybe we could have a `value/label` pattern here which is common where the `value` is the string to be stored in the DB that needs to never change, and then a `label` would be the human-readable string that could be changed to whatever in the future without having to worry about data integrity in the database (good for multiple language support as well)

otech47 (changes requested):

sofiarm21:
@otech47 changes requested reviewed, ready for merge

otech47 (approved):

Issue #289, opened 2021-02-09
Add UI for adding/editing client payments manually


otech47:
I think the Client Detail would be best but adding from the Project Payments could be convenient too

PR #316, opened 2021-02-16
Client add manual payment

sofiarm21:
### **Issue #289**

**Description:**

This pr contains the necessary changes to add and edit a payment directly from the app.

**Breakdown**

* The new page `PaymentsAddPage`, rendered on the route `/clients/:clientId/payments/add` contains the form to create the new payment
* The new mutation `CREATE_PAYMENT` stores the newly created payment in the DB/
* The component `PaymetnEditDialog` let the user edit the selected payment from the client page when the edit payment button is selected

**Proof of implementation**

https://www.loom.com/share/4ebdc41adc4b454db4f7e09ba4cd7e13

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -12,6 +12,7 @@ import ClientDetailPage from './pages/ClientDetailPage'
 import ClientsListPage from './pages/ClientsListPage'
 import HomePage from './pages/HomePage'
 import LoginPage from './pages/LoginPage'
+import PaymentsAddPage from './pages/PaymentsAddPage'
 import ProjectDetailPage from './pages/ProjectDetailPage'
 import ProjectsListPage from './pages/ProjectsListPage'
 
@@ -49,12 +50,17 @@ class App extends React.Component {
                         component={LoginPage}
                     />
                     <PrivateRoute
+                        exact
                         path='/clients/:clientId'
                         component={ClientDetailPage}
                     />
                     <PrivateRoute
                         path='/client/add'
                         component={AddClientPage}
+                    />
+                    <PrivateRoute
+                        path='/clients/:clientId/payments/add'
+                        component={PaymentsAddPage}
                     />
                     <PrivateRoute
                         path='/projects/:projectId'
--- a/src/components/PaymentEditDialog.js
+++ b/src/components/PaymentEditDialog.js
@@ -1,0 +1,139 @@
+import React, { useEffect, useState } from 'react'
+import {
+    Box,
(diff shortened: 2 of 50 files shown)


src/App.js
```diff
@@ -49,13 +50,18 @@ class App extends React.Component {
...
                         path='/clients/:clientId'
                         component={ClientDetailPage}
                     />
                     <PrivateRoute
                         path='/client/add'
                         component={AddClientPage}
                     />
+                    <PrivateRoute
+                        path='/clients/:clientId/payments/add'
+                        component={PaymentsAddPage}
```
otech47, line 63: Suggestion: Change to `AddPaymentPage` since it is both consistent with `AddClientPage` and also just makes a bit more sense when reading it

src/components/PaymentEditDialog.js
```diff
@@ -0,0 +1,201 @@
...
+        if (!dateIncurred || !paymentAmount) {
+            setDisableEdit(true)
+        } else if (
+            dateIncurred == moment(payment.date_incurred, 'x').format('YYYY-MM-DD') &&
+            datePaid == moment(payment.date_paid, 'x').format('YYYY-MM-DD')
+            && paymentAmount == payment.amount / 100) {
+            setDisableEdit(true)
+        } else {
+            setDisableEdit(false)
+        }
```
otech47, line 103: Got confused at this logic block because of the improper indent. Should be:

```
if (!dateIncurred || !paymentAmount) {
    setDisableEdit(true)
} else if (
    dateIncurred == moment(payment.date_incurred, 'x').format('YYYY-MM-DD') &&
    datePaid == moment(payment.date_paid, 'x').format('YYYY-MM-DD')
    && paymentAmount == payment.amount / 100
) {
    setDisableEdit(true)
} else {
    setDisableEdit(false)
}
```

Otherwise the function call looks like a 4th condition of the `else if`

src/pages/PaymentsAddPage.js
```diff
@@ -0,0 +1,44 @@
...
+                    alignItems='center'
+                />
+                <Grid item xs={8}>
+                    <Typography align='left'>
+                        <strong>
+                            {`Enter info below to add a payment`}
+                        </strong>
+                    </Typography>
+                    <Box mt={5}>
+                        <PaymentsAddForm clientId={clientId}/>
```
otech47, line 36: Suggestion: `AddPaymentForm` makes a bit more sense

PR #326, opened 2021-02-20
Total allocated by currency

sofiarm21:
### **Issue #301**

**Description:**

This pr contains the necessary changes for both backend and frontend, to be able to retrieve and show in the UI how much is been allocated to a contributor grouped by the currency.

**Breakdown**

* The contributor type `TotalAllocatedByCurrency` is an objet that contains the attributes `currency` and `amount`
* The `ContributorType` now has a new property called  `paid_by_currency` which is an array of object of `TotalAllocatedByCurrency` type  
* On the frontend side the function `renderPaidToContributorByCurrency` in the `ContributorInfoTile` maps and creates a list with the total paid by currency
* I started differing on the gql type which attributes are on the database and which are calculated leaving a comment `#The following attributes are calculated and aren't on the database`

**Proof of implementation**

Screen Shot 2021-02-22 at 10 23 24 AM
Screen Shot 2021-02-22 at 10 08 17 AM

### DIFF
--- a/api/schema/resolvers/ContributorResolver.js
+++ b/api/schema/resolvers/ContributorResolver.js
@@ -5,8 +5,8 @@ const toggl = require('../../handlers/toggl')
 
 module.exports = {
     Contributor: {
-        paid_by_currency: async (contributor, args, { models }) => {
-            const totalPaydByCurrencyQuery = await models.Client.findAll({
+        paidByCurrency: async (contributor, args, { models }) => {
+            const totalPaidByCurrencyQuery = await models.Client.findAll({
                 group: 'currency',
                 attributes: ['currency'],
                 raw: true,
@@ -23,16 +23,16 @@ module.exports = {
                     }
                 }
             })
-            totalPaydByCurrency = []
-            totalPaydByCurrencyQuery.map(t => {
-                totalPaydByCurrency.push(
+            totalPaidByCurrency = []
+            totalPaidByCurrencyQuery.map(t => {
+                totalPaidByCurrency.push(
                     {
                         currency: t['currency'],
                         amount: t['Projects.Allocations.amount']
                     }
                 )
             })
-            return totalPaydByCurrency
+            return totalPaidByCurrency
         },
         permissions: (contributor, args, { models }) => {
             return models.Permission.findAll({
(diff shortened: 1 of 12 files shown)


api/schema/resolvers/ContributorResolver.js
```diff
@@ -1,8 +1,39 @@
 const { AuthenticationError } = require('apollo-server')
+const { col, fn } = require('sequelize')
+
 const toggl = require('../../handlers/toggl')
 
 module.exports = {
     Contributor: {
+        paid_by_currency: async (contributor, args, { models }) => {
```
otech47, line 8: I'm not sure how strict this might be but I would recommend keeping consistent with custom properties as camelCase as `paidByCurrency` 

I use this pattern often and it is useful because when you see `snake_case` properties anywhere in the code you can have high confidence it is a column in the DB whereas if you see a `camelCase` property it is a clue that it is not a DB column and is instead a custom function or derived property

api/schema/resolvers/ContributorResolver.js
```diff
@@ -1,8 +1,39 @@
...
+            const totalPaydByCurrencyQuery = await models.Client.findAll({
```
otech47, line 9: Fix typo: `totalPaidByCurrencyQuery`

api/schema/resolvers/ContributorResolver.js
```diff
@@ -1,8 +1,39 @@
...
+                    include: {
+                        model: models.Allocation,
+                        attributes: [[fn('sum', col('amount')), 'amount']],
+                        where: {
+                            contributor_id: contributor.id
+                        },
+                    }
+                }
+            })
+            totalPaydByCurrency = []
```
otech47, line 26: Fix typo: `totalPaidByCurrency `

api/schema/types/ContributorType.js
```diff
@@ -15,6 +14,9 @@ module.exports = gql`
         permissions: [Permission]
         timeEntries: [TimeEntry]
         rates: [Rate]
+        #The following attributes are calculated and aren't on the database
```
otech47, line 17: Change to `"The following attributes are calculated and aren't on the database"` to expose this as documentation

otech47 (changes requested):

sofiarm21:
@otech47 changes requested resolved, ready for merge

Issue #311, opened 2021-02-16
Create empty states for contributor detail page


sofiarm21:
<img width="1440" alt="Screen Shot 2021-02-16 at 11 42 17 AM" src="https://user-images.githubusercontent.com/49292858/108085817-1cec3680-704c-11eb-94d9-93c7b700d174.png">

PR #329, opened 2021-02-22
Contributor page empty state

sofiarm21:
### **Issue #311**

**Description:**

This pr contains a very simple implementation, the empty states for the `ContributorDetailPage`

**Breakdown**

* I create a new component called `EmptyState` that takes a description and an icon classname (ideally from font awesome) as parameters and displays it as an empty state styled component. This component can be reused for all future empty states.

**Proof of implementation**

Screen Shot 2021-02-21 at 10 39 35 PM

### DIFF
--- a/src/components/ContributorAllocations.js
+++ b/src/components/ContributorAllocations.js
@@ -9,6 +9,7 @@ import {
 
 import AllocationAddForm from './AllocationAddForm'
 import AllocationTile from './AllocationTile'
+import EmptyState from './EmptyState'
 import LoadingProgress from './LoadingProgress'
 import { GET_CONTRIBUTOR_ALLOCATIONS, GET_CONTRIBUTOR_INFO } from '../operations/queries/ContributorQueries'
 import { white } from '../styles/colors.scss'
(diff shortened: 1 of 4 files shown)


src/components/ContributorAllocations.js
```diff
@@ -88,7 +89,16 @@ const ContributorAllocations = (props) => {
                 </Grid>
                 <Grid item xs={12}>
                     <Grid container spacing={5}>
-                        {renderAllocations({ allocations: contributorAllocations.allocations })}
+                        {
+                            contributorAllocations.allocations.length
```
otech47, line 93: This logic is confusing because it's not clear whether you are trying to render if there is more than 1 allocation in the array or if you are simply checking whether the value exists or is defined or is an array which would be better written more explicitly with `Array.isArray(contributorAllocations.allocations)`

Please resolve

src/components/ContributorProjectsCollab.js
```diff
@@ -56,11 +57,20 @@ const ContributorProjectsCollab = (props) => {
...
                     <Typography variant='h5' color='primary'>
                         <strong>
-                            {`Project Involvment`}
+                            {`Project Involvement`}
                         </strong>
                     </Typography>
                     <Grid container>
-                        {renderProjects({ projects: projects })}
+                        {
+                            projects.lenght
```
otech47, line 65: Same comment as above and also a typo here

otech47 (changes requested):

src/components/ContributorAllocations.js
sofiarm21: @otech47 I change this and since what I'm trying to verify is if the array contains elements now I use the lodash function `isEmpty` that does the same but it's more readable

sofiarm21:
@otech47 changes requested resolved and bug found fixed

PR #337, opened 2021-02-23
Project payments page

sofiarm21:
### **Issue #336**

**Description:**

This pr contains the necessary changes for the implementation of two elements on the `ProjectPaymentsPage`

1. The Project payments overview, is a section that shows 
    *  Total paid from the client (only the sum of the payments that has been confirmed)
    * Total allocated (confirmed and proposed)
    * Total profit
    * Total allocated proposed
   
2. The proposed section shows the total that has been proposed, and is expandable so also shows the allocations that conform to that amount.

For implementing all of this I had to make changes both in the backend and the frontend

**Breakdown:**

* new calculate attributes in the `ProjectType`.`totalAllocated` that returns the total allocated from a project, it has filters like `fromDate`, `toDate`, and `confirmedOnly` that will return only the allocated amount that has been confirmed (has a payment id associated). `totalIncurredPayments` that returns the sum of the payment amounts that the client hasn't paid yet
* On the `ProjectResolver` I made the necessary changes to resolve the new attributes returning values, both Int's
* On the `ClientQueries` the `GET_PROJECT_PAYMENTS_SUMMARY` get's all the info needed for the new `ProjectPaymentsSummary` component
* The new component `ProjectPaymentsSummary` mentioned before
* The `ProjectPoposedAllocationsTile` is the new component rendered that shows the total allocated and also has a dropdown with the details of the allocations.

**Proof of implementation:**

https://www.loom.com/share/b360ef8b7df9453ba5490780e40155ef

### DIFF
--- a/api/schema/types/ProjectType.js
+++ b/api/schema/types/ProjectType.js
@@ -48,10 +48,17 @@ module.exports = gql`
             fromDate: String
             toDate: String
         ): [TimeSpent]
-        totalPaid(
+        #calculated properties
+        totalAllocated(
             fromDate: String,
             toDate: String
         ): Int
+        totalPaid(
+            fromDate: String,
+            toDate: String
+        ): Int # The total paid from the client, takes start date and ed date as attributes
+        totalIncurredPayments: Int # The total incurred not paid from the client
+
     }
 
     type AverageIssueCost {
--- a/api/schema/types/ProjectType.js
+++ b/api/schema/types/ProjectType.js
@@ -45,20 +45,20 @@ module.exports = gql`
             contributor_id: Int
         ): TimeSpent
         timeSpentPerContributor(
-            fromDate: String
+            fromDate: String,
             toDate: String
         ): [TimeSpent]
         #calculated properties
         totalAllocated(
             fromDate: String,
-            toDate: String
+            toDate: String,
+            confirmedOnly: Boolean
         ): Int
         totalPaid(
             fromDate: String,
             toDate: String
         ): Int # The total paid from the client, takes start date and ed date as attributes
         totalIncurredPayments: Int # The total incurred not paid from the client
-
     }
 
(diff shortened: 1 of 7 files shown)


api/schema/types/ProjectType.js
```diff
@@ -45,13 +45,20 @@ module.exports = gql`
...
+        totalAllocated(
+            fromDate: String,
+            toDate: String,
+            confirmedOnly: Boolean
+        ): Int
         totalPaid(
             fromDate: String,
             toDate: String
-        ): Int
+        ): Int # The total paid from the client, takes start date and ed date as attributes
```
otech47, line 60: In GQL syntax it is actually better to expose these comments to the GraphQL explorer by writing as so:

```
"
"The total paid from the client, takes start date and end date as attributes"
totalPaid(
    fromDate: String,
    toDate: String
): Int
The total incurred not paid from the client
totalIncurredPayments: Int
```

Only use the `#` operator if you are putting comments in shouldn't show up in the GraphQL explorer

Also separately for this comment in particular the "takes start date and ed date as attributes" is redundant since it is already clear from the GQL specification that these are the arguments to pass

src/components/ProjectPayments.js
```diff
@@ -83,17 +96,34 @@ const ProjectPayments = (props) => {
...
+                <ProjectPaymentsSummary
+                    project={getProjectById}
+                    currencyInformation={currencyInformation}
+                />
+                {allocatedPayments.length != 0 &&
+                    <PaymentsList
+                        payments={payments}
+                        project={getProjectById}
+                    />
+                }
```
otech47, line 108: Proper use of indentation for conditional render 👍

api/schema/types/ProjectType.js
sofiarm21: resolved on pr #326

Issue #322, opened 2021-02-19
Add the ability to delete payments


otech47:
With confirmation prompt

PR #340, opened 2021-02-23
Delete payment functionality

sofiarm21:
### **Issue #322**

**Description:**

This pr contains the necessary changes to be able to delete a payment

**Important:**

Rn, when a payment is deleted, all the corresponding payments that were depending on this payment are also deleted by the SQL database, it has the configuration of _delete on cascade_

**Breakdown:**

* The new component `DeleteConfirmationDialog` is an entire UI component that is made having in mind that could be reused in multiple parts of the app, that's why I tried to make it as abstract as possible
* The `DeletePayment` component is more a logic component than a UI component, and it's the middle layer between the `PaymentTile` and the `DeleteConfirmationDialog`. Has the mutation to delete a payment given its `id`. Is the logic layer that customizes the `DeleteConfirmationDialog`
* The `PaymentTile` now renders a delete button if the app matches the route `/clients/:clientId`
* The `DELETE_PAYMENT ` apollo client mutation created on `PaymentMutations` takes the id of the payment to delete and returns the number of payments deleted

**Proof of implementation:**

https://www.loom.com/share/6accf1b1f5e44896b34a672407615bed

### DIFF
--- a/src/components/ClientPayments.js
+++ b/src/components/ClientPayments.js
@@ -26,8 +26,7 @@ const ClientPayments = (props) => {
                 <PaymentsList
                     payments={payments}
                 />
-            )
-            : (
+            ) : (
                 <PaymentsEmptyState/>
             )
     )
--- a/src/components/DeleteConfirmationDialog.js
+++ b/src/components/DeleteConfirmationDialog.js
@@ -1,6 +1,6 @@
 import React from 'react'
-
 import {
+    Box,
     Button,
     Dialog,
     DialogActions,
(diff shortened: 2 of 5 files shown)


src/components/ClientPayments.js
```diff
@@ -26,8 +26,7 @@ const ClientPayments = (props) => {
                 <PaymentsList
                     payments={payments}
                 />
-            )
-            : (
+            ) : (
```
otech47, line 29: Nice code styling correction 👍

otech47 (approved):
@sofiarm21 note minor code styling changes: https://github.com/setlife-network/trinary/pull/340/commits/9508777c11c3964e137d0a9d8e94c22b7f85e300

Issue #251, opened 2021-01-27
Add currency to Rates


otech47:
Just like we have it on the ClientType might as well refactor the other models to add a simple column as well as re-using the field on the [ClientAddForm](https://github.com/setlife-network/project-trinary/blob/develop/src/components/ClientAddForm.js) for creating/editing allocations with rates

PR #341, opened 2021-02-23
Rate currency

sofiarm21:
### **Issue #251**

**Description:**

This pr has the necessary implementations to add `currency` into `rates`. This pr has by changes in the DB model, changes in the backend, and changes in the frontend. The extension of this functionality includes adding the `currency` attribute in the DB model and being able to add the currency when creating a new allocation

**Breakdown**

* New migration already run on development/staging database called `addCurrenccyToRates`. This adds a new column `currency` of type `String` to the `Rates` table
* Added `currency` attribute into the `ProjectType`
* I implemented a currency dropdown selector in the `AllocationAddForm` as I did on the create client form.
* Now when creating an allocation by default the currency will be the currency of the client, if there's a past allocation for the contributor the currency will be the currency of the last allocation made to this contributor.
* Added the `currency` attribute to the queries made of the `rate`

**Notice**

Even that now each allocation will have its own currency everything is set to be populated with the currency of the client, this is because I haven't worked yet on the conversion between currencies. This can be done in other pr.

**Proof of implementation**

https://www.loom.com/share/9ad90b6f8a624eef851ca4b4938fdb4b

### DIFF
--- a/api/migrations/20210223183216-addCurrencyToRates.js
+++ b/api/migrations/20210223183216-addCurrencyToRates.js
@@ -1,0 +1,19 @@
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        return queryInterface.sequelize.transaction(t => {
+            return Promise.all([
+                queryInterface.addColumn('Rates', 'currency', {
+                    type: Sequelize.DataTypes.STRING
+                }, { transaction: t })
+            ])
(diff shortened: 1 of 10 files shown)


sofiarm21:
closes #251

otech47 (approved):

PR #343, opened 2021-02-25
Edit/Delete allocations

sofiarm21:
### **Issue #334**

**Description:**

This pr contains the necessary implementation to be able to delete/edit an allocation. Now each allocation can be selected and the allocation detail will be displayed. From this view, you can consult edit the information, edit the allocation or delete it. The allocation attributes that can be updated are the `payment_id`, `start_date`, `end_date`, the recalculated `amount`, and the `rate_id`. We can edit the allocations from the `ContributorTile` allocations dropdown, the `PaymentTile` allocations dropdown, and the `AllocationTile` on the `ContributorsPage`
 
**Breakdown:**

* The new component `AllocationOverview` is the container that holds the two parts of the allocation detail view, `EditAllocationInfo` and `EditAllocationRate`. Is the parent of these two components and handles all the inner logic of making the requests, updating, and deleting. Basically handles the logic while `EditAllocationInfo` and `EditAllocationRate` handles the UI, that's why exists bidirectional communication between these components and the parent. Only needs the allocation as the parameter, any other info will be query by the self component with the `allocation_id`
* The `handleUpdateAllocation` function in the `AllocationOverview` makes the update of the allocation, not the rate. For the rate, it looks if there exists a rate with the same values and if not creates a new one and update the allocation with the new values and the new `rate_id`.
* The mutation function `deleteAllocation` deletes the allocation given an `allocation_id` as a parameter but not the rate that goes with the allocation
* The `EditAllocationInfo` Shows the allocation info like the project, contributor, and client name, and has a dropdown to select a different payment from the one that the client has made. All this info is passed to the parent component for when the `Edit` button gets pressed.
* The `EditAllocationRate` contains more information about the allocation and its rate but also is displayed as a form so all the fields are editable. The component is very similar to the one that is used to create the allocations, and even reuses the components `RateProratedMonthlyForm` and `RateMaxBudgetForm` that are used in the `AllocationAddForm`. Works the same way as the `EditAllocationInfo` component, all its states are passed to the parent component.
* I added the necessary mutations and queries to store the data in the DB using Apollo Client. `DELETE_ALLOCATION` `UPDATE_ALLOCATION` both in `AllocationMutations.js`

**Note**

The issue https://github.com/setlife-network/trinary/issues/347 has to do with the edit allocation functionality but it will be resolved in another pr.

**Implementation proof**

https://www.loom.com/share/fa5415c7f44a48d6aa623dd180a20a78https://www.loom.com/share/fa5415c7f44a48d6aa623dd180a20a78

### DIFF
--- a/src/components/AllocationOverview.js
+++ b/src/components/AllocationOverview.js
@@ -13,6 +13,7 @@ import DeleteOutlinedIcon from '@material-ui/icons/DeleteOutlined'
 
 import DeleteConfirmationDialog from './DeleteConfirmationDialog'
 import EditAllocation from './EditAllocation'
+import LoadingProgress from './LoadingProgress'
 import { GET_ALLOCATION_INFO } from '../operations/queries/AllocationQueries'
 import { GET_PAYMENT_ALLOCATIONS } from '../operations/queries/PaymentQueries'
 import { GET_PROJECT_PAYMENTS } from '../operations/queries/ProjectQueries'
@@ -65,7 +66,7 @@ const AllocationOverview = (props) => {
         onClose()
     }
 
-    if (loadingAllocation) return 'Loading'
+    if (loadingAllocation) return <LoadingProgress/>
     if (errorAllocation) return `Error ${errorAllocation}`
 
     const { getAllocationById: allocation } = dataAllocation
--- a/src/components/AllocationOverview.js
+++ b/src/components/AllocationOverview.js
@@ -115,8 +115,6 @@ const AllocationOverview = (props) => {
 
     useEffect(() => {
         if (contributorAllocation) {
-            console.log('contributorAllocation');
-            console.log(contributorAllocation);
             getClientPayments({
                 variables: {
                     clientId: contributorAllocation.project.client.id
--- a/src/components/AllocationOverview.js
+++ b/src/components/AllocationOverview.js
@@ -1,0 +1,106 @@
+import React from 'react'
+import {
+    Box,
+    Dialog,
+    DialogTitle,
+    Grid,
+    Typography
+} from '@material-ui/core/'
+import moment from 'moment'
+
+import EditAllocation from './EditAllocation'
+import { formatAmount, selectCurrencyInformation } from '../scripts/selectors'
+
+const AllocationOverview = (props) => {
+
+    const {
+        allocation,
+        onClose,
+        open
+    } = props
+
+    console.log('allocation');
+    console.log(allocation);
+
+    const currencyInformation = selectCurrencyInformation({
+        currency: allocation.project.client.currency
+    })
+    const paymentAmount = formatAmount({
+        amount: allocation.payment.amount / 100,
+        currencyInformation: currencyInformation
+    })
+
+    /*
+    __typename: "Allocation"
+    amount: 320000
+    contributor: {__typename: "Contributor", id: 197, name: "Oscar Lafarga"}
+    end_date: "1617148800000"
+    id: 189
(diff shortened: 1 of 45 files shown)


src/components/PaymentTile.js
```diff
@@ -264,6 +269,13 @@ const PaymentTile = (props) => {
                 open={openEditPayment}
                 onClose={() => handleEditPayment(false)}
             />
+            {selectedAllocation &&
+                <AllocationOverview
+                    allocationInfo={selectedAllocation}
+                    onClose={() => handleAllocationClicked(false)}
+                    open={openAllocationOverview}
```
otech47, line 276: Suggestion:

```
onOpen={openAllocationOverview}
```

to be more consistent with `onClose`

src/components/RateProratedMonthlyForm.js
```diff
@@ -28,15 +28,16 @@ const RateProratedMonthlyForm = (props) => {
...
     const [totalWeeks, setTotalWeeks] = useState(null)
     const [totalHours, setTotalHours] = useState(0)
 
     useEffect(() => {
-        if (rateCurrency) {
-            setCurrency(rateCurrency)
+        if (!rateCurrency) {
+            setRateCurrency(clientCurrency)
         }
+        setCurrency(rateCurrency)
```
otech47, line 40: Not sure about these state structures dealing with `currency` maybe there is some refactoring/cleanup just leaving a note here but it's not clear exactly what's going on here, maybe even a comment to explain if this really is the best way to manage the currency state

src/components/ProjectProposedAllocationsTile.js
```diff
@@ -62,9 +72,9 @@ const ProjectProposedAllocationsTile = (props) => {
                             </Grid>
                             <Grid items xs={7}>
                                 <Typography color='secondary' variant='caption'>
-                                    {`${currencyInformation['symbol']}${a.rate.hourly_rate}/hr (
-                                    ${a.rate.type == 'monthly_rate' ? 'monthly rate' : 'max budget'}
-                                )`}
+                                    {`
+                                        ${currencyInformation['symbol']}${a.rate.hourly_rate}/hr (${a.rate.type == 'monthly_rate' ? 'monthly rate' : 'max budget'})
```
otech47, line 76: This line is too long and we should enforce some max columns in the linter at some point in the future to prevent spaghetti code

src/components/AllocationOverview.js
```diff
@@ -0,0 +1,276 @@
...
+        onCompleted: dataContributorRates => {
+            setContributorRates(dataContributorRates.getContributorById)
+        }
+    })
+
+    const [createRate, {
+        dataNewRate,
+        loadingNewRate,
+        errorNewRate
+    }] = useMutation(CREATE_RATE)
```
otech47, line 79: Code style suggestion

```
const [
    createRate,
    {
        dataNewRate,
        loadingNewRate,
        errorNewRate
    }
] = useMutation(CREATE_RATE)

```

src/components/AllocationOverview.js
```diff
@@ -0,0 +1,276 @@
...
+        } else {
+            //create rate
+            const newRate = await createRate({
+                variables: {
+                    hourly_rate: rate.hourly_rate.toString(),
+                    total_expected_hours: Number(rate.total_expected_hours),
+                    type: rate.type,
+                    currency: selectedCurrency,
+                    contributor_id: contributor.id
+                }
```
otech47, line 170: Just noting the inconsistency here:

```
{
    'hourly_rate': ...
}

{
    hourly_rate: ...
}
```

I would prefer without single quotes there's really no need for them

otech47 (approved):

PR #351, opened 2021-03-05
Minor cleanup and props simplification

otech47:
@sofiarm21 just had a few minutes here to show an example of the kind of code cleanup I like to do between releases just browsing a few random files


sofiarm21:
> @sofiarm21 just had a few minutes here to show an example of the kind of code cleanup I like to do between releases just browsing a few random files

@otech47 thanks for this, I'll a major cleanup after each release from now on

sofiarm21 (approved):

Issue #291, opened 2021-02-09
Add open and merged PRs as a secondary metric to ProjectIssues


otech47:
- Note "merged" PRs should be counted and not "closed" PRs since that would be a redundant metrics since that code may not have ever been merged and if it has then that's counted with merged PRs

otech47:
> * Note "merged" PRs should be counted and not "closed" PRs since that would be a redundant metrics since that code may not have ever been merged and if it has then that's counted with merged PRs

This issue overall was implemented but this note in the description was not considered.

Currently the logic will show all closed PRs but it is possible to close a PR without merging the code so that wouldn't be useful to count in these metrics. Will tackle this in a separate issue

PR #358, opened 2021-03-09
Pull requests as metrics and sync Github issues from UI

sofiarm21:
### **Issue #291 #359 #360**

**Description**

This pr resolves multiples issues listed above, the principal one was to add some metrics about the open and closed pull request on GitHub, like the metrics of the issues, for making this I had to make frontend and backend changes, both in this pr. Besides that, I fixed the bugs described in #359 and #360 

**Breakdown**

#291
* The `syncPullRequests ` function on the `dataSyncs` file, since Github returns the pull requests as issues I make use of the GitHub handler function `fetchRepoIssues ` to get all the issues and then I filter the pull requests.
* The `githubPullRequestsClosed` and `githubPullRequestsOpened` are attributes on the Contributor type that returns an integer, both accept as parameters `start date` and `end date`, but it's optional. On the contributor, I added the code to resolve these two attributes/functions
* On the `ProjectQueries` I add these two new attributes and I display them inside the `ProjectIssuesMetrics` component
* I created a button on the `ProjectIssues` component to be able to run the mutation `syncProjectIssues`  from the UI instead of having to make this sync from the apollo console
* The mutation `SYNC_GITHUB_ISSUES` called on the `ProjectIssues` component and declared on the `ProjectMutations`

#359 
* I fixed this bug inside the `syncGithubIssues` function on the `dataSyncs.js` file passing the right attribute `html_url` to the `findIssueByGithubUrl` function, instead of passing the `URL attribute. This is what was causing the bug.

#360 
* As I mentioned before GitHub takes the pull requests as issues, so the problem was that I wasn't filtering the pull request on the `syncGithubIssues` function causing that the pull requests were also being stored on the DB. On another pr, we could store both issues and PR's making a small change to the DB to differentiate them if we think it's convenient.
* To fix this I map the issues returned by the GitHub function `fetchRepoIssues` and store in a separate array all the issues that don't have the `pull_request` attribute

**Proof of implementation**

https://www.loom.com/share/e62d35a5818b40ea997b7438676ff593

Screen Shot 2021-03-10 at 9 17 20 AM
Screen Shot 2021-03-10 at 9 18 12 AM

**Additional note**
The API key that shows as deleted on one of the tests files is an old key don't worry about it being on the source control

### DIFF
--- a/src/operations/mutations/ProjectMutations.js
+++ b/src/operations/mutations/ProjectMutations.js
@@ -18,6 +18,25 @@ export const ADD_PROJECT = gql`
     }
 `
 
+export const SYNC_GITHUB_ISSUES = gql`
+    mutation syncGithubIssues($project_id: Int!) {
+        syncProjectIssues(project_id: $project_id) {
+            id
+        }
+    }
+`
+
+export const SYNC_PROJECT_GITHUB_CONTRIBUTORS = gql`
+    mutation syncProjectGithubContributors($project_id: Int!, $github_personal_key: String){
+        syncProjectGithubContributors(project_id: $project_id, github_personal_key: $github_personal_key) {
+            id
+            name
+            github_id
+            github_handle
+        }
+    }
+ `
+
 export const UPDATE_PROJECT = gql`
     mutation updateProjectById($project_id: Int!, $date: String!, $expected_budget:Int!, $name: String!, $github_url: String, $expected_budget_timeframe: String, $toggl_url: String){
         updateProjectById(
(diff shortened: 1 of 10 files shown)


src/operations/mutations/ProjectMutations.js
```diff
@@ -18,6 +18,25 @@ export const ADD_PROJECT = gql`
...
+    mutation syncGithubIssues($project_id: Int!) {
+        syncProjectIssues(project_id: $project_id) {
+            id
+        }
+    }
+`
+
+export const SYNC_PROJECT_GITHUB_CONTRIBUTORS = gql`
+    mutation syncProjectGithubContributors($project_id: Int!, $github_personal_key: String){
+        syncProjectGithubContributors(project_id: $project_id, github_personal_key: $github_personal_key) {
```
otech47, line 31: @sofiarm21 not sure what the impact is of this but I noticed in the ProjectType we have `github_personal_key` defined but in the ProjectResolvers we have `args.githubPersonalKey` so this discrepancy will fail every time but since it has a fallback (https://github.com/setlife-network/trinary/blob/b376789381faf2b47a1ea9854a4f4826d4288b8d/api/schema/resolvers/ProjectResolver.js#L556) it probably doesn't actually cause an error.

Merging for now but as a separate issue it's probably worth doing a project-wide search for `githubPersonalKey` just to refactor/remove after making sure it is needed and works properly in the cases it should be

otech47 (approved):

Issue #203, opened 2020-12-31
Tweak TimeEntry model to allow description field


sofiarm21:
For a future version, we should implement a list with all the time entries with the description of each, for this we need to add a string attribute on the TimeEntries table to store the description attribute that comes from toggle with each time entry

PR #364, opened 2021-03-12
Toggl description attribute

sofiarm21:
### **Issue #203**

**Description:**

This pr contains the necessary backend changes to add a new column `description` of type `String` into the `TimeEntries` table, using migrations.

**Breakdown**

* `addDescriptionToTimeEntries` file contains the Sequelize migration to add the new column `description`. I made both, the up and the down migration. in the model `TimeEntry` file I declared the new attribute of type `string` and can be nullable
* In the file `timeLogging` i made some changes to the `addTimeEntries` script to add the description of the time entry into the DB

**Implementation proof**

https://www.loom.com/share/de58dd5c352a4f8b9a2d0fd2395dcd8d

### DIFF
--- a/api/models/TimeEntry.js
+++ b/api/models/TimeEntry.js
@@ -23,10 +23,6 @@ module.exports = (sequelize) => {
             allowNull: false,
             unique: true
         },
-        description: {
-            type: DataTypes.STRING,
-            allowNull: false
-        },
         start_time: {
             type: DataTypes.DATE,
             allowNull: false
@@ -46,6 +42,9 @@ module.exports = (sequelize) => {
                 model: 'Projects',
                 key: 'id',
             }
+        },
+        description: {
+            type: DataTypes.STRING
         }
     },
     {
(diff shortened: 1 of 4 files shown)


api/models/TimeEntry.js
```diff
@@ -23,6 +23,10 @@ module.exports = (sequelize) => {
             allowNull: false,
             unique: true
         },
+        description: {
+            type: DataTypes.STRING,
+            allowNull: false
```
otech47, line 28: @sofiarm21 This should not be a required field right? Also this doesn't match what's in the video since in there it shows Allow Null is true in the database which is the ideal structure I think. Also the TimeEntries that don't come up with a description seem blank in the database in the video when it's probably better if they are left as NULL since we would now kind of have a mix of null/empty states so better to just have NULL instead of an empty string

otech47 (changes requested):

api/models/TimeEntry.js
sofiarm21: @otech47 I resolved this on the most recent commit, now the descriptions that comes as blank from toggl are stored as NULL.

Issue #323, opened 2021-02-19
Implement a local configuration of the database


otech47:

sofiarm21:
@otech47 is this document how create a local database and connect it to Trinary? If I'm not wrong what's needed is create the db from an sql client and change the env variables, or do you have something else in mind?

otech47:
@mecoccaro please document the problems you ran into setting up and update the README to help future developers trying to set up the project and database

PR #381, opened 2021-03-29
Implement a local configuration of the database

mecoccaro:
Issue #323 

Added to the package.json file an extra script for Windows users, so they can run the project without having to modify the normal `npm run server`. Modified the README file with the new instructions for Windows users and added instructions for use of a local database.

Implementation proof

https://user-images.githubusercontent.com/36824674/112883704-a146c600-909c-11eb-8c7f-08cad70e7262.mp4

Local DB
imagen

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,14 +35,8 @@ npm run ui
 
 4. Open a new Terminal tab and run the web server
 
-If running on MacOS
 ```
 npm run server
-```
-
-If running on Windows
-```
-npm run server-windows
 ```
 
 5. UI should open in the browser at `localhost:6002`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -35,14 +35,36 @@ npm run ui
...
+npm run server-windows
+```
+
 5. UI should open in the browser at `localhost:6002`
 
 If using VSCode, be sure to disable the Prettier: Format on Save feature to prevent unwanted changes to the code styles.
 
+## Using a local database 
+
+If you want to set a local databe follow this steps:
```
otech47, line 54: Fix typo: `set up a local database, follow these steps:`

README.md
```diff
@@ -35,14 +35,36 @@ npm run ui
...
+
+1. Download MySQL server (Usually it comes with others features).
```
otech47, line 56: Maybe provide a link here to help the reader find the correct version of MySQL to download?

README.md
```diff
@@ -35,14 +35,36 @@ npm run ui
...
+
+2. Configure your environment.
+
+3. Create a connection.
+
+4. Create a schema.
```
otech47, line 62: These 3 steps are a bit too general for someone inexperienced. Does this involve running commands? It's nice and convenient to provide some sample commands for the reader to be able to run on their machine to make sure this step completed properly

package.json
```diff
@@ -57,6 +57,7 @@
   "scripts": {
     "ui": "cross-env PORT=6002 react-scripts start",
     "server": "PORT=6001 node server",
+    "server-windows": "SET PORT=6001 && node server.js",
```
otech47, line 60: Okay solution for now but I think even better would be a `cross-env` implementation (https://www.npmjs.com/package/cross-env)

README.md
```diff
@@ -35,14 +35,36 @@ npm run ui
...
+
+5. In your `.env` file set the MYSQL variables with your own values.
```
otech47, line 64: I think this is the 1st time a `.env` file is mentioned in the README. Let's try and add some more detail here and [minimize required knowledge](https://github.com/setlife-network/trinary/blob/develop/CONTRIBUTING.md#1-minimize-required-knowledge)

otech47 (changes requested):

package.json
mecoccaro: I did not know about that implementation, it works perfectly! Thanks.

README.md
```diff
@@ -43,6 +43,47 @@ npm run server
 
 If using VSCode, be sure to disable the Prettier: Format on Save feature to prevent unwanted changes to the code styles.
 
+## Using a local database
```
otech47, line 46: 🔥

otech47 (approved):

Issue #378, opened 2021-03-20
Add client_id into the permissions table


sofiarm21:
Int nullable
The contributor that creates the client is the only one that has `write` access to that client

PR #382, opened 2021-03-29
Restricted access to clients by client_id (permissions table)

sofiarm21:
### **Issue #378**

**Description:** 

This pr contains the necessary changes to add a `client_id` column in the `Permissions table` to be able to control the permission level of the contributors. For now, the default permission is `write` access to the contributor that created the client

**Breakdown:** 

* The column was added via a new migration
* The `client_id` is of type Integer and could be nullable
* When a new client is created the permission is automatically created with the contributor_id and the `client_id`. The `contributor_id` is given via the cookies session, but in case we are adding the client directly from the Apollo server console, we have to pass the `contributor_id` as one of the parameters.

**Implementation proof:**

https://www.loom.com/share/980318d98496485e9dd768fc779df14b

**Note:**

We have to decide who's going to have `read` permission to the client, there are two options

1. All the contributors have `read` access to all the clients.
2. Only the contributors thas has `read` or `write` permission to at least one project owned by the client will have `read` access to that client. This can be a script that gets called every time a user logs in.

Once we decided which one of these will be I will open a new issue to implement it.

### DIFF
--- a/api/schema/resolvers/ClientResolver.js
+++ b/api/schema/resolvers/ClientResolver.js
@@ -53,11 +53,12 @@ module.exports = {
     Mutation: {
         createClient: async (root, { createFields }, { cookies, models }) => {
             try {
+                if (!cookies.userSession || createFields.contributor_id) {
+                    throw new Error('A contributor id is required');
+                }
                 const newlyCreatedClient = await models.Client.create({
                     ...createFields
                 })
-                console.log('createFields');
-                console.log(createFields);
                 const contributor = (
                     await models.Contributor.findByPk(
                         cookies ? cookies.userSession : createFields.contributor_id
--- a/api/schema/resolvers/ClientResolver.js
+++ b/api/schema/resolvers/ClientResolver.js
@@ -49,10 +49,25 @@ module.exports = {
         }
     },
     Mutation: {
-        createClient: (root, { createFields }, { models }) => {
-            return models.Client.create({
+        createClient: async (root, { createFields }, { cookies, models }) => {
+            const newlyCreatedClient = await models.Client.create({
                 ...createFields
             })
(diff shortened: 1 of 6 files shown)


sofiarm21:
closes #378

api/schema/types/ClientType.js
```diff
@@ -16,11 +16,19 @@ module.exports = gql`
...
         email: String,
         currency: String,
+        contributor_id: Int
+    }
+
+    input ClientInput {
+        is_active: Boolean,
+        name: String,
+        email: String,
+        currency: String
```
otech47, line 31: I think we don't need 2 separate input types especially since there are no strictly required`!` fields

otech47 (changes requested):

api/schema/types/ClientType.js
sofiarm21: The CinetCreateInput actually should be requires, so I will leave both input objects but one of them will be required

api/schema/resolvers/ClientResolver.js
```diff
@@ -49,10 +49,25 @@ module.exports = {
...
     Mutation: {
-        createClient: (root, { createFields }, { models }) => {
-            return models.Client.create({
+        createClient: async (root, { createFields }, { cookies, models }) => {
+            const newlyCreatedClient = await models.Client.create({
                 ...createFields
             })
+            const contributor = (
+                await models.Contributor.findByPk(
+                    cookies ? cookies.userSession : createFields.client_id
```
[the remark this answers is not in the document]
sofiarm21: fixed using `try catch` and making a validation before creating the client

sofiarm21:
@otech47 change srequested resolved

PR #383, opened 2021-03-31
Disable zoom "double-tap" gesture on mobile devices

sofiarm21:


sofiarm21:
After some research, I found the possible solution [here](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action#manipulation) but I don't know how to simulate a double-tap gesture on the mobile simulated with chrome dev tools, so cannot tell for sure.

otech47 (approved):

Issue #371, opened 2021-03-16
Add this/last week/month/year options to time tracking date filter


otech47:
a la Toggl

![Screen Shot 2021-03-16 at 10 55 54 AM](https://user-images.githubusercontent.com/4914611/111330289-2fa35c80-8646-11eb-9638-492b1586f594.png)

PR #390, opened 2021-04-13
Date filter selector and date range selector polish

sofiarm21:
### **Issue #371**

**Description:**

This pr implements multiple range selectors for the range date selector on the project overview page to filter the time entries. Also, I made a new UI for this component.

**Breakdown:**

* The `TIME_RANGES` object in the constants file contains the different date range selection options. The `description` is what the user sees (the label). The `periodRange` has the string that corresponds to the momentjs library. The `since` is a number that represents how long ago this period id, 0 means is the actual time (week / 0 -> this week) (week / 1 -> last week) (week / 2 -> 2 weeks ago) (year / 1 -> last year). I did this because this way I can use one single implementation and function to make the filtering, so we can add as many ranges as we want without implementing a function for each.
* The function `renderFixedRangeTimes` in `ProjectTimeTracking` using moment and a simple logic converts the `TIME_RANGES` in actual date ranges.
* The new component `RangeDatePickerInput` isolates the UI for the date range selector into its own component so we don't have to repeat the same code every time we want to add a date range selector

**Implementation proof:**

https://www.loom.com/share/5b896ebe4ac145a7833dea8d727ab499

Screen Shot 2021-04-13 at 2 23 32 PM

### DIFF
--- a/src/components/ProjectTimeTracking.js
+++ b/src/components/ProjectTimeTracking.js
@@ -1,7 +1,7 @@
 import React, { useState } from 'react'
 import { useLazyQuery, useQuery } from '@apollo/client'
 // import DatePicker from 'react-datepicker'
-// import 'react-datepicker/dist/react-datepicker.css'
+import 'react-datepicker/dist/react-datepicker.css'
 import { RangeDatePicker } from 'react-google-flight-datepicker'
 import 'react-google-flight-datepicker/dist/main.css'
 import {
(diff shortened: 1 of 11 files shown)


src/styles/RangeDatePicker.scss
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+    .react-datepicker__day--in-selecting-range {
+        background: $setlife-blue;
+        &:hover {
+            background: $setlife-blue;
+        }
+    }
+    .react-datepicker__day--keyboard-selected {
+        background: $setlife-blue;
```
otech47, line 26: Just noticed a variable inconsistency `$setlife-blue` in kebab-case and `$light_grey` in snake_case so just making a note to capture this maybe in a code cleanup/refactor issue for the future

src/constants/index.js
```diff
@@ -116,3 +116,30 @@ export const NAV_TITLES = [
     }
 ]
 export const SMALL_LOGO_URL = 'https://project-trinary.s3.amazonaws.com/images/SmallLogo.png'
+export const TIME_RANGES = [
+    {
+        description: 'This week',
+        periodRange: 'weeks',
+        since: 0
+    }, {
```
otech47, line 124: NIT: open objects on a separate line

```
...
},
{
...
```

src/components/ProjectTimeTracking.js
```diff
@@ -40,19 +51,51 @@ const ProjectTimeTracking = (props) => {
...
             } })
         }
     }
 
+    const setFixedRangedTime = ({ startDate, endDate }) => {
+        setStartDate(startDate['_d'])
+        setEndDate(endDate['_d'])
+    }
+
+    const renderFixedRangeTimes = (timeRanges) => {
```
otech47, line 75: NIT: `setFixedTimeRange` and `renderFixedTimeRanges` reads better

Issue #372, opened 2021-03-16
Fix date timezone mismatching when creating allocations


otech47:
- Created an allocation from October 1st to October 31st
- Checked the allocation on Edit Allocation from the ProjectPayments page and the date now shows 9/30 to 10/30

otech47:
Something similar is still happening here:

https://www.loom.com/share/685f2a8e80c74d83952041c6a0645050

Not sure if we should re-open this or create a new issue @sofiarm21 but maybe it's the same fix in a different component?

PR #392, opened 2021-04-14
Date timezone fix

sofiarm21:
### **Issue #372**

**Description:**

This pr fix a bug that was causing on date transformation from timestamp to human readable date a difference of a day when we transformed to the local date. I fixed this by not transforming the dates that are coming from the DB to local time because they are already stored in local time, so no need to do it again

**Breakdown:**

* It was fixed simple changing the moment formatting from `moment.(timestamp, x). format()` to `moment.utc(timestamp, 'x').format()`
* I changed most of the moment date conversions, but if there's any left, this would be the correct way to format it
* I also fixed some bugs that I found making testing

**Implementation proof**

https://www.loom.com/share/88051ef342d74f79b46fbbfc0cae7455

### DIFF
--- a/src/components/ContributorAllocations.js
+++ b/src/components/ContributorAllocations.js
@@ -98,7 +98,7 @@ const ContributorAllocations = (props) => {
                             : (
                                 <EmptyState
                                     description='This contributor has no allocations at the moment'
-                                    iconClassname='fas fa-money-check'
+                                    iconClassName='fas fa-money-check'
                                 />
                             )
                         }
--- a/src/components/ContributorAllocations.js
+++ b/src/components/ContributorAllocations.js
@@ -93,7 +93,7 @@ const ContributorAllocations = (props) => {
                 </Grid>
                 <Grid item xs={12}>
                     <Grid container spacing={5}>
-                        {isEmpty(contributorAllocations.allocations)
+                        {!isEmpty(contributorAllocations.allocations)
                             ? renderAllocations({ allocations: contributorAllocations.allocations })
                             : (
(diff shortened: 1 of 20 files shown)


src/components/ContributorAllocations.js
```diff
@@ -93,15 +93,14 @@ const ContributorAllocations = (props) => {
...
-                                        description='This contributor has no allocations at the moment'
-                                        iconClassname='fas fa-money-check'
-                                    />
-                                )
+                        {isEmpty(contributorAllocations.allocations)
+                            ? renderAllocations({ allocations: contributorAllocations.allocations })
+                            : (
+                                <EmptyState
+                                    description='This contributor has no allocations at the moment'
+                                    iconClassname='fas fa-money-check'
```
otech47, line 101: Fix camelcasing: `iconClassName`

src/components/ContributorAllocations.js
```diff
@@ -93,15 +93,14 @@ const ContributorAllocations = (props) => {
...
-                                ? renderAllocations({ allocations: contributorAllocations.allocations })
-                                : (
-                                    <EmptyState
-                                        description='This contributor has no allocations at the moment'
-                                        iconClassname='fas fa-money-check'
-                                    />
-                                )
+                        {isEmpty(contributorAllocations.allocations)
+                            ? renderAllocations({ allocations: contributorAllocations.allocations })
+                            : (
```
otech47, line 98: Test multi-line PR comments

otech47 (approved):

Issue #393, opened 2021-04-14
Create a webhook endpoint for syncing Stripe Customers with Trinary Clients


otech47:
- Set up API endpoint
- Fetch all Stripe customers
- Check DB to see if each Stripe Customer already has a client with a matching `external_uuid`
- If none exists, create the Client with name, email, currency (if provided), external_uuid
- If it already exists, update the Client name & email

PR #407, opened 2021-04-29
Sync Stripe customers

mecoccaro:
**Issue #393**

**Description**

In this Pull Request the following is implemented:

* When the webhook `customer.created` and `customer.updated` its triggered to the endpoint `/api/webhooks/clients`  it creates in the DB the new client or it updates the client, according to the type key.

**Breakdown**

* Initially the endpoint for this webhook was  `/api/webhooks/clients/created` or `/api/webhooks/clients/updated`  but the webhook json file has a "type" key that distinguishes what type of hook it is:
```
 "type": "customer.updated",
```
So with this key you can compare between a created client or updated one.
* When a new client its created, the `external_uuid` row is filled with the customer id from Stripe, the same happends with an updated client, if it does not have the `external_uuid` that its trying to update it creates the new client with the `external_uuid`

**Implementation proof**

https://user-images.githubusercontent.com/36824674/117694794-8aac8800-b18d-11eb-84c7-f4208151001d.mp4

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,7 +32,7 @@ module.exports = {
         'jsx-quotes': ['error', 'prefer-single'],
         'keyword-spacing': ['error', { 'after': true }],
         'lines-between-class-members': 'off',
-        //'linebreak-style': ['error', 'off'],
+        'linebreak-style': ['error', 'windows'], //For windows user using ESlint.
         'max-len': 'off',
         'new-cap': 'off',
         'no-bitwise': 'off',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,7 +32,7 @@ module.exports = {
         'jsx-quotes': ['error', 'prefer-single'],
         'keyword-spacing': ['error', { 'after': true }],
         'lines-between-class-members': 'off',
-        'linebreak-style': ['error', 'windows'], //For windows user using ESlint.
+        //'linebreak-style': ['error', 'off'],
         'max-len': 'off',
         'new-cap': 'off',
         'no-bitwise': 'off',
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -19,7 +19,29 @@ const automations = module.exports = (() => {
                 external_uuid: clientInformation.external_uuid
             })
         }
+    }
 
+    const updateClient = async (params) => {
+        const clientToUpdate = await db.models.Client.findOne({
+            where: {
+                external_uuid: params.clientInformation.external_uuid
+            }
+        })
+        if (clientToUpdate) {
+            return db.models.Client.update({
+                email: params.clientInformation.email,
+                currency: params.clientInformation.currency,
+                name: params.clientInformation.name,
+                updated_at: moment(params.clientInformation.date_created),
+                external_uuid: params.clientInformation.external_uuid
+            }, {
+                where: {
+                    id: clientToUpdate.id
+                }
+            })
+        } else {
+            createClient({ clientInformation: params.clientInformation })
+        }
     }
 
     const createPayment = async ({ paymentInformation }) => {
@@ -139,6 +161,7 @@ const automations = module.exports = (() => {
 
     return {
         createClient,
+        updateClient,
         createPayment,
         getUserOrganizations,
         getOrganizationRepos,
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -6,6 +6,22 @@ const db = require('../models')
 
 const automations = module.exports = (() => {
 
+    const createClient = async ({ clientInformation }) => {
+        const client = await getClientFromExternalId({ id: clientInformation.external_uuid })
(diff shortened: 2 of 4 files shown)


server.js
```diff
@@ -105,11 +106,108 @@ app.get('/api/oauth-redirect', (req, res) => { //redirects to the url configured
...
+        console.log(`An error ocurred: ${err}`)
+    }
+})
+
 app.use('/api/graph/v/:vid', express.json(), (req, res, next) => {
     console.log(`Incoming API v${req.params.vid} request on worker PID ${process.pid}`)
     next()
 })
 
+app.post('/api/webhooks/clients',express.json(), (req, res, next) => {
```
otech47, line 182: Add space after: `/clients', express.json(`

server.js
```diff
@@ -105,11 +106,108 @@ app.get('/api/oauth-redirect', (req, res) => { //redirects to the url configured
...
+    const body = req.body.data.object
+    const clientInformation = {
+        email: body.email,
+        currency: body.currency,
+        name: body.name,
+        date_created: req.body.created,
+        external_uuid: body.id
+    }
+    const webhookType = req.body.type
+    if (webhookType === "customer.created"){
```
otech47, line 192: Always use single quotes

api/modules/automations.js
```diff
@@ -1,8 +1,105 @@
...
+                    email: params.clientInformation.email
+                }
+            })
+        }
+        if (clientToUpdate) {
+            return db.models.Client.update({
+                email: params.clientInformation.email,
+                currency: params.clientInformation.currency,
+                name: params.clientInformation.name,
+                updated_at: moment(params.clientInformation.date_created),
```
otech47, line 48: `updated_at` should be a current timestamp instead of the date created from Stripe

- Check if removing this causes Sequelize to update this automatically, if not use moment to get the current timestamp

api/modules/automations.js
```diff
@@ -1,8 +1,105 @@
...
+        }
+        if (params.clientInformation.email && (clientToUpdate === null)){
+            clientToUpdate = await db.models.Client.findOne({
+                where: {
+                    email: params.clientInformation.email
+                }
+            })
+        }
+        if (clientToUpdate) {
+            return db.models.Client.update({
```
otech47, line 44: Alternative:

```
clientToUpdate.email: = params.clientInformation.email
clientToUpdate.save()
```

otech47 (changes requested):

otech47 (changes requested):
@mecoccaro if you could just make sure your linter is set up in your IDE and then go through each file to check that all linting errors are resolved then this will be good to merge 👍

server.js
```diff
@@ -160,6 +160,36 @@ app.post('/api/webhooks/payment_intent/succeeded', (req, res) => {
     }
 })
 
+app.post('/api/webhooks/clients', express.json(), (req, res, next) => {
```
otech47, line 163: Test if `express.json()` can be removed

server.js
```diff
@@ -160,6 +160,36 @@ app.post('/api/webhooks/payment_intent/succeeded', (req, res) => {
...
+    }
+    const webhookType = req.body.type
+    if (webhookType === 'customer.created'){
+        apiModules.automations.createClient({ clientInformation })
+            .then(() => {
+                res.sendStatus(200)
+            })
+            .catch(err => {
+                console.log(`An error ocurred: ${err}`)
+            })
```
otech47, line 181: Prefer async/await over promises

```
try {
    const createdClient = await apiModules.automations.createClient({ clientInformation })
    res.sendStatus(200)
} catch (error) {
    console.log(`An error ocurred: ${err}`)
}
```

otech47 (changes requested):
Resolve remaining linting errors

.eslintrc.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
         'jsx-quotes': ['error', 'prefer-single'],
         'keyword-spacing': ['error', { 'after': true }],
         'lines-between-class-members': 'off',
-        //'linebreak-style': ['error', 'off'],
+        'linebreak-style': ['error', 'windows'], //For windows user using ESlint.
```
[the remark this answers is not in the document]
mecoccaro: Sure, I will open a PR for that.

server.js
```diff
@@ -105,11 +106,108 @@ app.get('/api/oauth-redirect', (req, res) => { //redirects to the url configured
...
 
+app.post('/api/webhooks/clients',express.json(), (req, res, next) => {
+    const body = req.body.data.object
+    const clientInformation = {
+        email: body.email,
+        currency: body.currency,
+        name: body.name,
+        date_created: req.body.created,
+        external_uuid: body.id
+    }
```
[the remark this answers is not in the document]
mecoccaro: In this case Destructuring won't work because we call an object from the response of stripe.

Issue #409, opened 2021-05-14
Build a Stripe webhook for deleting draft invoices


otech47:
Linked payments in Trinary should also be deleted if the draft invoice is deleted on Stripe

PR #432, opened 2021-06-17
Webhook for deleting Draft Invoices

mecoccaro:
**Issue #409**
**Description**

A new webhook is created for deleting draft invoices from Stripe. When a draft invoice is deleted from the Stripe dashboard, it activates the webhook and goes the endpoint that was created for deleting the invoice with the external_uuid.

**Implementation proof**

https://user-images.githubusercontent.com/36824674/122448807-25f60180-cf73-11eb-94cb-c52d524e5ff5.mp4

### DIFF
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -20,6 +20,20 @@ const automations = module.exports = (() => {
         }
     }
 
+    const deleteDraftInvoicesFromStripe = (params) => {
+        let deletedInvoice
+        try {
+            deletedInvoice = db.models.Payment.destroy({
+                where: {
+                    external_uuid: params.invoiceId
+                }
+            })
+        } catch (err) {
+            console.log(`An error ocurred while deleting invoice: ${err}`)
+        }
+        return deletedInvoice
+    }
+
     const getClientFromEmail = (params) => {
         return db.models.Client.findOne({
             where: {
@@ -146,6 +160,7 @@ const automations = module.exports = (() => {
 
     return {
         createPayment,
+        deleteDraftInvoicesFromStripe,
         getClientWithExternalId,
         getUserOrganizations,
         getOrganizationRepos,
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -29,7 +29,7 @@ const automations = module.exports = (() => {
                 }
             })
         } catch (err) {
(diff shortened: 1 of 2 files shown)


api/modules/automations.js
```diff
@@ -20,6 +20,20 @@ const automations = module.exports = (() => {
         }
     }
 
+    const deleteDraftInvoicesFromStripe = (params) => {
+        let deletedInvoice
```
otech47, line 24: Change to `deletedPayment`

server.js
```diff
@@ -143,6 +143,15 @@ app.post('/api/webhooks/invoice/updated', (req, res) => {
         res.send('payment not ready to allocate')
     }
 })
+app.post('/api/webhooks/invoice/delete', async (req, res) => {
+    const invoiceId = req.body.data.object.id
+    try {
+        await apiModules.automations.deleteDraftInvoicesFromStripe({ invoiceId })
```
otech47, line 149: Refactor this to call the function in a new module called `/api/modules/budgeting.js`

- sort index.js A-Z with new module
- Move and rename `deleteDraftInvoicesFromStripe` to `deletePaymentByStripeInvoiceId`

api/modules/automations.js
```diff
@@ -20,6 +20,20 @@ const automations = module.exports = (() => {
...
+        try {
+            deletedInvoice = db.models.Payment.destroy({
+                where: {
+                    external_uuid: params.invoiceId
+                }
+            })
+        } catch (err) {
+            console.log(`An error ocurred: ${err}`)
+        }
```
otech47, line 33: Let's tweak this to address some issues:

- If the payment does exist but it failed to delete for some reason, we want to send back a status of 400 (check what Stripe expects in webhook responses) to make sure Stripe tries again when possible
- If the payment does not exist, then we need to make sure Stripe does not keep trying

otech47 (changes requested):

otech47:
Merging now since the comments above will be addressed in #435

api/modules/automations.js
mecoccaro: If we send another code different than 2xx, stripe will keep trying several times until it stops and marks the event as failure.
Changing so we send a status of 400 in case of failure.
More details on https://stripe.com/docs/webhooks/build

Issue #397, opened 2021-04-14
Change the empty state of the Clients page to guide the user on Stripe synchronization


otech47:
Before the first client is created, there should be some instructions on how the Stripe integrated client-customer sync works

## Tasks

- [x] Test backend functionalities to make sure all features still work as expected without any Stripe credentials
- [x] Provide a way for the frontend UI to know whether Stripe credentials are valid
- [x] Show instructions on how to use Stripe for client management (if valid Stripe credentials are detected)
- [x] Show instructions on how to deploy/configure the app to enable the Stripe integration

See https://github.com/setlife-network/trinary/blob/develop/src/components/ClientsEmptyState.js

### Stripe not detected copy

```
To integrate with Stripe for client management, please configure the application with valid Stripe credentials (see example environment configuration here https://github.com/setlife-network/trinary/blob/develop/.env.example#L13)

Please see the Stripe documentation (https://stripe.com/docs/keys) to generate valid credentials
```

### Stripe detected copy

```
Creating or updating a Stripe customer will trigger a synchronized Client

Any updates to the name and email of a Client will reflect in your Stripe Dashboard
```

otech47:
Copy for this will be provided by @otech47

PR #436, opened 2021-06-24
Empty state of clients and Stripe Credentials

mecoccaro:
**Issue #397**
**Description**

Before the first client is created on Trinary, we show a message on how the Stripe integrated client-customer sync works, first we check if the project has the Stripe credentials set it into the env file, if not we explain how to set the credentials on the file, otherwise we explain that clients can be managed from Stripe Customers.

**Implementation proof**

https://user-images.githubusercontent.com/36824674/123306670-95c63800-d4ef-11eb-9717-a35b6b1574e0.mp4

### DIFF
--- a/src/components/ClientsList.js
+++ b/src/components/ClientsList.js
@@ -13,7 +13,7 @@ import { HAS_VALID_STRIPE_CREDENTIALS } from '../operations/queries/ConfigQuerie
 
 const ClientsList = (props) => {
     const history = useHistory()
-
+    const { loading: loadingS, error: errorS, data: dataStripe } = useQuery(HAS_VALID_STRIPE_CREDENTIALS);
     const { loading, error, data } = useQuery(GET_CLIENTS);
 
     const renderClientTiles = (clients) => {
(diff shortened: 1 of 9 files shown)


src/components/ClientsList.js
```diff
@@ -6,12 +6,14 @@ import { orderBy } from 'lodash'
...
 import ClientsEmptyState from './ClientsEmptyState'
+import ClientsEmptyStateStripe from './ClientsEmptyStateStripe'
 import LoadingProgress from './LoadingProgress'
 import { GET_CLIENTS } from '../operations/queries/ClientQueries'
+import { HAS_VALID_STRIPE_CREDENTIALS } from '../operations/queries/ConfigQueries';
 
 const ClientsList = (props) => {
     const history = useHistory()
-
+    const { loading: loadingS, error: errorS, data: dataStripe } = useQuery(HAS_VALID_STRIPE_CREDENTIALS);
```
otech47, line 16: Suggestion

```
const {
    data: hasValidStripeCredentialsData
} = useQuery(HAS_VALID_STRIPE_CREDENTIALS);
```

otech47 (approved):

Issue #435, opened 2021-06-23
Refactor webhook functions to corresponding module


otech47:
Expanding on this: https://github.com/setlife-network/trinary/pull/432#discussion_r657316396_

We want to move some functionality to improve modularity
- [x] Resolve all comments in: https://github.com/setlife-network/trinary/pull/432
- [x] webhook `/invoice/updated` / `updatePaymentFromStripe` to `budgeting` module
- [x] Rename `updatePaymentFromStripe` to `updatePaymentByStripeInvoiceId`
- [x] webhook `/invoice/paid` / `updateDatePaidPayment` to `budgeting` module
- [x] Create new functions as needed to keep our `server.js` file as a thin routing layer with minimal functional logic

PR #444, opened 2021-06-30
Refactoring webhook functions to corresponding module

mecoccaro:
**Issue #432 and #435**
**Description**

Cleaning up the server.js file so all the webhook functions are moved to its corresponding module, in this case the two modules for the existing webhooks are clientManagement and Budgeting. There are also some quick refactoring on variables that were set with var and changed to const.

- [x] Resolve all comments in: https://github.com/setlife-network/trinary/pull/432
- [x] webhook `/invoice/updated` / `updatePaymentFromStripe` to `budgeting` module
- [x] Rename `updatePaymentFromStripe` to `updatePaymentByStripeInvoiceId`
- [x] webhook `/invoice/paid` / `updateDatePaidPayment` to `budgeting` module
- [x] Create new functions as needed to keep our `server.js` file as a thin routing layer with minimal functional logic

### DIFF
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -61,7 +61,6 @@ const automations = module.exports = (() => {
     }
 
     return {
-        getClientWithExternalId,
         getUserOrganizations,
         getOrganizationRepos,
     }
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -3,14 +3,6 @@ const db = require('../models')
 
 const automations = module.exports = (() => {
 
-    const getClientWithExternalId = (params) => {
-        return db.models.Client.findOne({
-            where: {
-                external_uuid: params.id
-            }
-        })
-    }
-
     const getUserOrganizations = async (params) => {
         //user organizations are the organizations that the contributor is added as a internal collaborator
         //we also add the self github contributor user for implementation details
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -20,20 +20,6 @@ const automations = module.exports = (() => {
         }
     }
 
-    const deleteDraftInvoicesFromStripe = (params) => {
-        let deletedInvoice
-        try {
-            deletedInvoice = db.models.Payment.destroy({
-                where: {
-                    external_uuid: params.invoiceId
-                }
-            })
-        } catch (err) {
-            console.log(`An error ocurred: ${err}`)
-        }
-        return deletedInvoice
-    }
-
     const getClientFromEmail = (params) => {
         return db.models.Client.findOne({
             where: {
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -1,39 +1,13 @@
-const moment = require('moment')
-
 const github = require('../handlers/github')
-const { GITHUB } = require('../config/credentials')
 const db = require('../models')
 
 const automations = module.exports = (() => {
-
-    const getClientFromEmail = (params) => {
-        return db.models.Client.findOne({
-            where: {
-                email: params.email
-            }
-        })
-    }
 
     const getClientWithExternalId = (params) => {
         return db.models.Client.findOne({
             where: {
                 external_uuid: params.id
             }
-        })
-    }
-
-    const getPaymentWithExternalId = (params) => {
-        return db.models.Payment.findOne({
-            where: {
-                external_uuid: params.id
-            },
-            raw: true,
-        })
-    }
-
-    const getPaymentWithId = (params) => {
-        return db.models.Payment.findByPk(params.id, {
-            raw: true
         })
     }
 
@@ -95,7 +69,6 @@ const automations = module.exports = (() => {
     }
 
     return {
-        createPayment,
         getClientWithExternalId,
         getUserOrganizations,
         getOrganizationRepos,
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -6,20 +6,6 @@ const db = require('../models')
 
 const automations = module.exports = (() => {
 
-    const createPayment = async ({ paymentInformation }) => {
-        const client = await getClientWithExternalId({ id: paymentInformation.customer_id })
-        if (client) {
-            return db.models.Payment.create({
-                amount: paymentInformation.amount,
-                external_uuid: paymentInformation.external_uuid,
-                date_incurred: moment(paymentInformation.date_incurred['_d']).format('YYYY-MM-DD HH:mm:ss'),
-                date_paid: paymentInformation.date_paid ? moment(paymentInformation.date_paid['_d']) : null,
-                client_id: client.id,
-                external_uuid_type: paymentInformation.external_uuid_type
-            })
-        }
-    }
-
     const getClientFromEmail = (params) => {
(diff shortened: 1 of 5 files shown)


api/modules/budgeting.js
```diff
@@ -0,0 +1,103 @@
+const db = require('../models');
+const moment = require('moment')
+
+const budgeting = module.exports = (() => {
+
+    const createPayment = async ({ paymentInformation }) => {
+        const client = await getClientWithExternalId({ id: paymentInformation.customer_id })
```
otech47, line 7: `getClientWithExternalId` seems undefined here

api/modules/budgeting.js
```diff
@@ -0,0 +1,103 @@
...
+        if (client) {
+            return db.models.Payment.create({
+                amount: paymentInformation.amount,
+                external_uuid: paymentInformation.external_uuid,
+                date_incurred: moment(paymentInformation.date_incurred['_d']).format('YYYY-MM-DD HH:mm:ss'),
+                date_paid: paymentInformation.date_paid ? moment(paymentInformation.date_paid['_d']) : null,
+                client_id: client.id,
+                external_uuid_type: paymentInformation.external_uuid_type
+            })
```
otech47, line 16: Suggestion: deconstruct above to reduce repeated references to `paymentInformation`

```
const {
    amount,
    external_uuid,
    date_incurred,
    date_paid,
    amount,
} = paymentInformation
    
return db.models.Payment.create({
    amount: amount,
    external_uuid: external_uuid,
    date_incurred: moment(date_incurred['_d']).format('YYYY-MM-DD HH:mm:ss'),
    date_paid: date_paid ? moment(date_paid['_d']) : null,
    client_id: client.id,
    external_uuid_type: external_uuid_type
})
```

which of course then simplifies to

```
return db.models.Payment.create({
    amount,
    external_uuid,
    date_incurred: moment(date_incurred['_d']).format('YYYY-MM-DD HH:mm:ss'),
    date_paid: date_paid ? moment(date_paid['_d']) : null,
    client_id: client.id,
    external_uuid_type
})
```

api/modules/budgeting.js
```diff
@@ -0,0 +1,103 @@
...
+
+const budgeting = module.exports = (() => {
+
+    const createPayment = async ({ paymentInformation }) => {
+        const client = await getClientWithExternalId({ id: paymentInformation.customer_id })
+        if (client) {
+            return db.models.Payment.create({
+                amount: paymentInformation.amount,
+                external_uuid: paymentInformation.external_uuid,
+                date_incurred: moment(paymentInformation.date_incurred['_d']).format('YYYY-MM-DD HH:mm:ss'),
```
otech47, line 12: What is the `_d` notation here? It seems confusing and not obvious what it does or where it comes from.

If it's possible to remove it away that'd be best but if not a comment should be left so the reader is not confused why it is there

api/modules/clientManagement.js
```diff
@@ -29,30 +36,38 @@ const clientManagement = module.exports = (() => {
...
+                    email: clientInformation.email
                 }
             })
         }
         if (clientToUpdate) {
-            clientToUpdate.email = params.clientInformation.email
-            clientToUpdate.currency = params.clientInformation.currency
-            clientToUpdate.name = params.clientInformation.name
-            clientToUpdate.external_uuid = params.clientInformation.external_uuid
+            clientToUpdate.email = clientInformation.email
```
otech47, line 64: Suggestion: deconstruct `clientInformation` to avoid many references to it

api/modules/budgeting.js
```diff
@@ -0,0 +1,103 @@
...
+        await db.models.Payment.update({
+            date_paid: moment(paymentToUpdate.date_paid, 'YYYY-MM-DD')
+        }, {
+            where: {
+                id: paymentToUpdate.id
+            }
+        })
+    }
+
+    const updatePaymentByStripeInvoiceId = async ({ data }) => {
```
otech47, line 71: Remember to [avoid this kind of naming convention](https://github.com/setlife-network/trinary/blob/develop/CONTRIBUTING.md#4-avoid-redundant-programming-terminology) since it is obvious there is "data" being handled, but here we waste the opportunity to add context to what kind of data it is

otech47 (changes requested):

api/modules/budgeting.js
mecoccaro: @sofiarm21 Can you help us here? so we can find the best solution!

sofiarm21: @mecoccaro yeah so `['_d']` notation is used to access the instance of the Date object that backs the moment object. So in this case, `date_incurred` is a moment object and we are accessing its internal property `_d`. We shouldn't be accessing internals like this and instead, we should be using the functions that the moment library provides to us.  
So instead of `moment(paymentInformation.date_incurred['_d']).format('YYYY-MM-DD HH:mm:ss')` this could be refactored to something like this `paymentInformation.date_incurred.format('YYYY-MM-DD HH:mm:ss')` 
Ik there are some places around the app that we are accessing the internals of the moment object but this should be refactored at some point. 

[Here](https://momentjs.com/guides/) is the documentation about the moment library in case you haven't checked it out already.

api/modules/automations.js
```diff
@@ -122,50 +60,10 @@ const automations = module.exports = (() => {
...
-        } else {
-            //the payment is not in the db, proceed to store it
-            createPayment({ paymentInformation: params.paymentInformation })
-        }
-    }
-
     return {
-        createPayment,
-        deleteDraftInvoicesFromStripe,
         getClientWithExternalId,
```
sofiarm21, line 64: @mecoccaro you should either delete this line or bring back the `getClientWithExternalId` to the file because currently, you are exporting a function that doesn't exist and is causing the server to fail on mount

api/modules/budgeting.js
```diff
@@ -0,0 +1,113 @@
...
+            date_paid: moment(paymentToUpdate.date_paid, 'YYYY-MM-DD')
+        }, {
+            where: {
+                id: paymentToUpdate.id
+            }
+        })
+    }
+
+    const updatePaymentByStripeInvoiceId = async ({ paymentObjectPayload }) => {
+        const datePaidOverride = paymentObjectPayload.custom_fields[findIndex(paymentObjectPayload.custom_fields, { 'name': 'date_paid' })]
```
sofiarm21, line 82: @mecoccaro As a recommendation, I would try to make the lines of code more concise to avoid spaghetti code, you could either break this code into multiple lines 
```
const datePaidOverride = paymentObjectPayload.custom_fields[
   findIndex(
      paymentObjectPayload.custom_fields, 
      { 'name': 'date_paid' }
   )
]
```
or create a constant with a significant name and store the value of `findIndex(paymentObjectPayload.custom_fields, { 'name': 'date_paid' })` in it and then make the reference with that value

otech47: Also `'name'` as a property name we should not use single quotes around it so we don't confuse it for a string. so it'd be just `{ name: 'date_paid' }`

otech47 (approved):

Issue #402, opened 2021-04-16
Create Stripe Invoices when a Payment is added manually


otech47:
- Should we create the Stripe Invoice first and let the webhook create the Trinary Payment? 
- A placeholder product ID will be used to create new Prices for the amount the user submits for the Payment
<img width="1843" alt="Screen Shot 2021-04-16 at 11 42 16 AM" src="https://user-images.githubusercontent.com/4914611/115056952-0e35ca80-9ea9-11eb-8422-a6da72a950ab.png">

otech47:
Placeholder product ID: https://dashboard.stripe.com/test/products/prod_JJXofAMeGR4HIS

sofiarm21:
<img width="835" alt="Screen Shot 2021-07-09 at 12 13 32 AM" src="https://user-images.githubusercontent.com/49292858/125022355-82b47b80-e04a-11eb-8231-ab670bc2e458.png">

PR #447, opened 2021-07-09
Stripe manual payment

sofiarm21:
### **Issue #402** 

**Description:**

This issue makes it possible to create a stripe invoice automatically when a trinary invoice is created manually from the web app. 

* The product id it's hardcoded to be `prod_JJXofAMeGR4HIS` which is the trinary placeholder product and the collection_method is also hardcoded to be `charge_automatically`. 
* Depending on the existence of `date_paid` the invoice will be automatically finalized or not. 
* The stripe invoice id is being added to the DB along with the rest of the payment record.
* Important to notice that in stripe the created and finalized dates will be different from the ones on trinary since are the ones that when the transaction is being made

**Important:**

Stripe doesn't permit different invoices of the same client with different currencies, so we should some way block the edit functionality for clients connected to stripe or create a new stripe client when this happens. This is the error that Stripe returns
Screen Shot 2021-07-09 at 1 08 11 AM

**Breakdown:**

* The`paymentManagement` module orchestrates all the stripes functionality regarding the payments and adds additional logic to that abstraction layer

**Implementation proof:**

https://www.loom.com/share/4c92a205f5af47a19a4ce86cf135d8f1

### DIFF
--- a/api/config/constants.js
+++ b/api/config/constants.js
@@ -28,5 +28,6 @@ module.exports = {
             github_permission_level: 'admin',
             project_permission_level: 'write'
         },
-    ]
+    ],
+    STRIPE_PRODUCT_PLACEHOLDER_ID: 'prod_JJXofAMeGR4HIS'
 }
--- a/api/config/constants.js
+++ b/api/config/constants.js
@@ -29,5 +29,5 @@ module.exports = {
             project_permission_level: 'write'
         },
     ],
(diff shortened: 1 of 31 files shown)


otech47 (approved):
@sofiarm21 
https://www.loom.com/share/e55b0c8e2cc54e0fac43fa54d78e4dbf

Issue #413, opened 2021-05-18
Add the ability to mark active/inactive clients


otech47:
- Add a "Show inactive clients" button below the active clients list on the `/home/clients` page. Clicking this button displays an inactive clients section below that. If the page is refreshed only the `ShowInactiveClientsButton` remains. If there are no inactive clients do not show the `ShowInactiveClientsButton`
- Add the ability to toggle the `active` property on clients from the EditClient page

otech47:
https://material-ui.com/components/switches/

otech47:
https://www.loom.com/share/ae5c4ab48fd344fa92717c491a348426

PR #465, opened 2021-07-16
Inactive/Active clients on Home and Edit Client

mecoccaro:
**Issue #413**
**Description**

On the home view we now have the ability to show Inactive Clients, if there are no inactive clients the option won't show. 
On the edit client view, we have the ability to edit the `is_active` status of a client.

Additionally, on the home view was fixed that it used to show all clients, no matter if they were active or inactive. Now they are separated.

### DIFF
--- a/api/schema/types/ClientType.js
+++ b/api/schema/types/ClientType.js
@@ -35,6 +35,7 @@ module.exports = gql`
     type Query {
         getClientById(id: Int!): Client
         getClients: [Client]
+        getInactiveClients: [Client]
         getActiveClientsCount: Int!
         getInactiveClientsCount: Int!
     }
--- a/api/schema/types/ClientType.js
+++ b/api/schema/types/ClientType.js
@@ -35,6 +35,7 @@ module.exports = gql`
     type Query {
         getClientById(id: Int!): Client
         getClients: [Client]
+        getActiveClients: [Client]
         getInactiveClients: [Client]
         getActiveClientsCount: Int!
         getInactiveClientsCount: Int!
--- a/api/schema/types/ClientType.js
+++ b/api/schema/types/ClientType.js
@@ -36,6 +36,7 @@ module.exports = gql`
         getClientById(id: Int!): Client
         getClients: [Client]
         getActiveClientsCount: Int!
+        getInactiveClientsCount: Int!
     }
 
     type Mutation {
--- a/src/components/ClientEditDialog.js
+++ b/src/components/ClientEditDialog.js
@@ -10,11 +10,13 @@ import {
     Grid,
     MenuItem,
     Select,
-    TextField
+    TextField,
+    FormGroup,
 } from '@material-ui/core/'
 
 import { UPDATE_CLIENT } from '../operations/mutations/ClientMutations'
 import { CURRENCIES } from '../constants'
+import Switch from '@material-ui/core/Switch';
 
 const ClientEditDialog = (props) => {
 
@@ -29,6 +31,7 @@ const ClientEditDialog = (props) => {
     const [clientName, setClientName] = useState(client.name)
     const [clientEmail, setClientEmail] = useState(client.email)
     const [clientCurrency, setClientCurrency] = useState(client.currency)
+    const [clientIsActive, setClientIsActive] = useState(client.is_active)
     const [disableEdit, setDisableEdit] = useState(true)
 
     const onEdit = async () => {
@@ -41,6 +44,9 @@ const ClientEditDialog = (props) => {
         }
         if (clientCurrency) {
             clientInfoToEdit['currency'] = clientCurrency
+        }
+        if (clientIsActive) {
+            clientInfoToEdit['is_active'] = clientIsActive
         }
         updateClient({
             variables: clientInfoToEdit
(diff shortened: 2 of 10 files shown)


api/schema/types/ClientType.js
```diff
@@ -35,7 +35,10 @@ module.exports = gql`
     type Query {
         getClientById(id: Int!): Client
         getClients: [Client]
+        getActiveClients: [Client]
+        getInactiveClients: [Client]
         getActiveClientsCount: Int!
+        getInactiveClientsCount: Int!
```
otech47, line 41: Suggestion:

An optimization here might look like this:

```
getClients(active: Boolean): [Client]
getTotalClients(active: Boolean): Int!
```

which might result in some LOC savings

Does not need to be changed now but an idea for #474

src/components/ClientEditDialog.js
```diff
@@ -10,11 +10,13 @@ import {
...
     MenuItem,
     Select,
-    TextField
+    TextField,
+    FormGroup,
 } from '@material-ui/core/'
 
 import { UPDATE_CLIENT } from '../operations/mutations/ClientMutations'
 import { CURRENCIES } from '../constants'
+import Switch from '@material-ui/core/Switch';
```
otech47, line 19: This import should be grouped together with the others

src/components/ClientEditDialog.js
```diff
@@ -120,6 +140,22 @@ const ClientEditDialog = (props) => {
...
                         </Grid>
+                        <Grid item xs={12} lg={5}>
+                            <Box my={2}>
+                                <Grid component='label' container alignItems='center' spacing={1}>
+                                    <Grid item>Inactive</Grid>
+                                    <Grid item>
+                                        <Switch
+                                            checked={state.checkedA}
+                                            onChange={handleChange}
+                                            name='checkedA'
```
otech47, line 151: what does the `A` stand for in `checkedA`? This name can probably be more descriptive like `activeSelector` or `isActiveCheckBox` or `clientActiveSwitch`

src/components/InactiveClientListManager.js
```diff
@@ -0,0 +1,37 @@
...
+import Switch from '@material-ui/core/Switch';
+
+import LoadingProgress from './LoadingProgress'
+import { GET_INACTIVE_CLIENTS_COUNT } from '../operations/queries/ClientQueries'
+import InactiveClientsList from './InactiveClientsList'
+
+const InactiveClientListManager = (props) => {
+
+    const { loading, error, data } = useQuery(GET_INACTIVE_CLIENTS_COUNT);
+    console.log(data)
```
otech47, line 16: Remove log

src/components/InactiveClientListManager.js
```diff
@@ -0,0 +1,37 @@
...
+            mb={3}
+            mx={1}
+            className='ClientListManager'
+        >
+            {
+                data.getInactiveClientsCount != 0
+                    ? (
+                        <InactiveClientsList />
+                    ) : ( false )
+            }
```
otech47, line 30: NIT Suggestion: Change to

```
{data.getInactiveClientsCount != 0
     ? <InactiveClientsList />
     : null
}
```

or 

```
{
     data.getInactiveClientsCount != 0
          ? <InactiveClientsList />
          : null
}
```

src/components/InactiveClientsList.js
```diff
@@ -0,0 +1,103 @@
...
+const InactiveClientsList = (props) => {
+    const history = useHistory()
+    const { loading, error, data } = useQuery(GET_INACTIVE_CLIENTS);
+
+    const handleChange = (event) => {
+        setState({ ...state, [event.target.name]: event.target.checked })
+    }
+
+    const [state, setState] = React.useState({
+        checkedA: false,
```
otech47, line 24: Rename to something more descriptive

src/components/InactiveClientsList.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+    const { loading: loadingInactive , error: errorInactive, data: dataInactive } = useQuery(GET_INACTIVE_CLIENTS_COUNT);
```
otech47, line 26: Change to

```
const {
    loading: loadingInactive,
    error: errorInactive,
    data: dataInactive
} = useQuery(GET_INACTIVE_CLIENTS_COUNT)
```

src/components/InactiveClientsList.js
```diff
@@ -0,0 +1,103 @@
...
+import LoadingProgress from './LoadingProgress'
+import {GET_INACTIVE_CLIENTS, GET_INACTIVE_CLIENTS_COUNT} from '../operations/queries/ClientQueries'
+import FormGroup from '@material-ui/core/FormGroup'
+import FormControlLabel from '@material-ui/core/FormControlLabel'
+import Switch from '@material-ui/core/Switch'
+import Box from '@material-ui/core/Box'
+
+const InactiveClientsList = (props) => {
+    const history = useHistory()
+    const { loading, error, data } = useQuery(GET_INACTIVE_CLIENTS);
```
otech47, line 17: Since there are 2 sets of `loading/error/data` maybe it would be better to be explicit in both places

```
{
    loading: loadingInactiveClients,
    error: errorInactiveClients,
    data: dataInactiveClients,
}
{
    loading: loadingInactiveClientsCount,
    error: errorInactiveClientsCount,
    data: dataInactiveClientsCount,
}
```

src/components/InactiveClientsList.js
```diff
@@ -0,0 +1,103 @@
...
+                                </Box>
+                            </Grid>
+                        ) : (
+                            false
+                        )
+                }
+                <Grid container>
+                    {
+                        state.checkedA ? clients.length != 0 ? renderClientTiles(clients) : false : false
+                    }
```
otech47, line 96: Nested ternaries like this are always tricky to make sure they are not confusing to the reader, I recommend:

```
{
    state.checkedA
        ? clients.length != 0
                ? renderClientTiles(clients)
                : null
        : null
}
```

Note you should use `null` instead of `false` to render nothing in React

otech47 (changes requested):

otech47:
@mecoccaro remember that testing instructions are required for PRs. Videos are easiest (Loom.com is a good option) but screenshots are okay as well

otech47 (approved):

Issue #453, opened 2021-07-13
Many create/update/delete functions do not reflect in UI until you refresh the page


otech47:
Tried with:

- Deleting allocations
- Creating clients and clicking the Home logo in nav
- Creating projects and clicking the Home logo in nav

There might be several other places so we should test all create/update/delete functions throughout the entire UI and make sure it updates the local state and does not require a refresh

Ovifer13:
https://github.com/trojanowski/react-apollo-hooks/issues/158

Ovifer13:
- [ ] Creating Clients.
- [ ] Creating Projects.
- [ ] Deleting Allocations.
- [ ] Creating a contributor is not updating Payments.

PR #472, opened 2021-07-19
UI refresh after adding, editing or deleting data.

Ovifer13:
**Issue #453**
Now when we add a client, project, or allocation the UI will update without having to refresh.

The problem was caused because when there is an _useQuery_  with Apollo, there is a [fetch policy](https://www.apollographql.com/docs/react/data/queries/#setting-a-fetch-policy), by default this policy will only check the cache and show only the store data on it without querying the GraphSQL server. To fix this issue we have to specify what policy do we want for the query.

For example:
```
const { loading, error, data } = useQuery(GET_DOGS, {
  fetchPolicy: "cache-and-network"
});
```

If we use _cache-and-network_ policy 
> Apollo Client executes the full query against both the cache and your GraphQL server. The query automatically updates if the result of the server-side query modifies cached fields.

### DIFF
--- a/src/components/ClientProjectsList.js
+++ b/src/components/ClientProjectsList.js
@@ -14,6 +14,7 @@ const ClientProjectsList = ({
 }) => {
 
     const { loading, error, data, networkStatus } = useQuery(GET_CLIENT_INFO, {
+        fetchPolicy: "cache-and-network",
         variables: { id: Number(clientId) }
     })
 
--- a/src/components/ClientProjectsList.js
+++ b/src/components/ClientProjectsList.js
@@ -14,7 +14,7 @@ const ClientProjectsList = ({
 }) => {
 
     const { loading, error, data, networkStatus } = useQuery(GET_CLIENT_INFO, {
-        fetchPolicy: "cache-and-network",
+        fetchPolicy: 'cache-and-network',
         variables: { id: Number(clientId) }
     })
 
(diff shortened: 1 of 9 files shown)


src/components/ClientProjectsList.js
```diff
@@ -14,6 +14,7 @@ const ClientProjectsList = ({
 }) => {
 
     const { loading, error, data, networkStatus } = useQuery(GET_CLIENT_INFO, {
+        fetchPolicy: "cache-and-network",
```
otech47, line 17: @Ovifer13 be sure to check your linter is properly set up I think @mecoccaro might have some recent experience with this if you have trouble

Double quote strings should be caught by the linter but always use single quotes of tildes 

```
'cache-and-network'
`cache-and-network`
```

otech47, line 17: @Ovifer13 be sure to check your linter is properly set up I think @mecoccaro might have some recent experience with this if you have trouble

Double quote strings should be caught by the linter but always use single quotes of tildes 

```
'cache-and-network'
`cache-and-network`
```

otech47 (changes requested):

src/components/ClientProjectsList.js
Ovifer13: Already set up linter and it is working perfectly!

otech47 (approved):

Issue #411, opened 2021-05-18
Fix various allocation bugs


otech47:
https://www.loom.com/share/ce9b825216cd462f9b3b6460c0dcd005

- [x] Sort payments to select from in an allocation by Date Incurred (most recent first)
- [x] When adding an allocation it says $0.00 allocated until refreshing
- [x] Can't edit the hourly rate of an existing allocation (edit button is disabled)
- [x] When trying to edit an allocation the Hours per week calculations do not update properly (in add allocation they do update properly)
- [x] Delete button click listening area is overlapping on top of Edit, should be moved it over to the right

(not shown in video)
- [x] When deleting an allocation from the Project Payments page, it does not disappear from the list until after refreshing the page
- [x] When opening a payment on the Project Payments page to edit an allocation, it is not clear that the allocation can be clicked on. Improve this UI with some indication that you can edit the allocation as well as the payment with separate buttons?

Ovifer13:
Task 2: _When adding an allocation it says $0.00 allocated until refreshing_ was already fixed on Issue #453

PR #476, opened 2021-07-22
Allocation bugs fixes

Ovifer13:
**Issue #411**
When you add an allocation, payments are sorted by **Date Incurred**
image

There is a new **Edit Button** in allocations instead of having to click the allocation to edit it.
I would like some feedback on this one if the button looks good in there or if it should be placed somewhere else.
image

The Edit Button will be only disabled if the total amount or the hourly rate is less or equals 0
The Edit and Delete buttons now have space in between and are responsive.
**Computer size**
image

**Mobile Size**
image

### DIFF
--- a/src/components/AllocationAddForm.js
+++ b/src/components/AllocationAddForm.js
@@ -341,7 +341,7 @@ const AllocationAddForm = (props) => {
             ? dataProjectContributors.allocations
             : null
     const payments = dataClientPayments
-        ? [...dataClientPayments.getProjectById.client.payments, { amount: null, date_paid: null }]
+        ? [...dataClientPayments.getProjectById.client.payments, { amount: null, date_paid: null }].sort((a, b) => b.date_incurred - a.date_incurred)
         : [null]
     const clientCurrency = (
         dataClientPayments
--- a/src/components/AllocationAddForm.js
+++ b/src/components/AllocationAddForm.js
@@ -341,7 +341,10 @@ const AllocationAddForm = (props) => {
             ? dataProjectContributors.allocations
             : null
     const payments = dataClientPayments
-        ? [...dataClientPayments.getProjectById.client.payments, { amount: null, date_paid: null }].sort((a, b) => b.date_incurred - a.date_incurred)
+        ? [...dataClientPayments.getProjectById.client.payments, { 
+            amount: null,
+            date_paid: null 
+        }].sort((a, b) => b.date_incurred - a.date_incurred)
         : [null]
     const clientCurrency = (
         dataClientPayments
--- a/src/components/AllocationOverview.js
+++ b/src/components/AllocationOverview.js
@@ -225,7 +225,8 @@ const AllocationOverview = (props) => {
     }
     
     const editButtonDisabled = (
-        updatedAllocationRate.total_amount <= 0 
+        updatedAllocationRate.total_amount <= 0 ||
+        updatedAllocationRate.hourly_rate <= 0
     )
 
     return (
--- a/src/components/AllocationOverview.js
+++ b/src/components/AllocationOverview.js
@@ -213,9 +213,11 @@ const AllocationOverview = (props) => {
     if (errorAllocation || errorClientPayments) return `Error`
 
     const { getAllocationById: allocation } = dataAllocation
-    const payments = [null]
-    if (clientPayments) {
-        payments.unshift(...clientPayments.payments)
+    const payments = dataClientPayments
+        ? [...dataClientPayments.getClientById.payments, { amount: null, date_paid: null }].sort((a, b) => b.date_incurred - a.date_incurred)
+        : [null]
+    if (payments) {
+        payments.pop();
     }
 
     if (!contributorAllocation) {
@@ -223,8 +225,7 @@ const AllocationOverview = (props) => {
     }
     
     const editButtonDisabled = (
-        updatedAllocationRate.total_amount == allocationInfo.amount &&
-        updatedAllocationRate.hourly_rate == allocationInfo.rate.hourly_rate
+        updatedAllocationRate.total_amount <= 0 
     )
 
     return (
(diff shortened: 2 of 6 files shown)


src/components/AllocationAddForm.js
```diff
@@ -341,7 +341,7 @@ const AllocationAddForm = (props) => {
             ? dataProjectContributors.allocations
             : null
     const payments = dataClientPayments
-        ? [...dataClientPayments.getProjectById.client.payments, { amount: null, date_paid: null }]
+        ? [...dataClientPayments.getProjectById.client.payments, { amount: null, date_paid: null }].sort((a, b) => b.date_incurred - a.date_incurred)
```
sofiarm21, line 344: This is considered spaghetti code, try to avoid it whenever you can by breaking it into multiple lines like this
```
? [...dataClientPayments.getProjectById.client.payments, { 
     amount: null, 
     date_paid: null 
}].sort((a, b) => b.date_incurred - a.date_incurred)
```
so it can be more readable

src/components/AllocationOverview.js
```diff
@@ -213,18 +213,20 @@ const AllocationOverview = (props) => {
     if (errorAllocation || errorClientPayments) return `Error`
 
     const { getAllocationById: allocation } = dataAllocation
-    const payments = [null]
-    if (clientPayments) {
-        payments.unshift(...clientPayments.payments)
+    const payments = dataClientPayments
+        ? [...dataClientPayments.getClientById.payments].sort((a, b) => b.date_incurred - a.date_incurred)
```
sofiarm21, line 217: The same comment of spaghetti code applies here

src/components/AllocationOverview.js
```diff
@@ -213,18 +213,20 @@ const AllocationOverview = (props) => {
...
 
     if (!contributorAllocation) {
         setContributorAllocation(allocation)
     }
     
     const editButtonDisabled = (
-        updatedAllocationRate.total_amount == allocationInfo.amount &&
-        updatedAllocationRate.hourly_rate == allocationInfo.rate.hourly_rate
+        updatedAllocationRate.total_amount <= 0 ||
+        updatedAllocationRate.hourly_rate <= 0
```
sofiarm21, line 229: 1. We should allow hourly rates of zero, so we should only check for negative numbers
2. I would go ahead and try a different implementation to make this logic more robust. Instead of adding logic to check the total_amount and  hourly_rate in here, we shouldn't let the user write negative values or letters at the text fields in the first place, so go deeper into the components of the `RateProratedMonthlyForm` and `RateMaxBudgetForm` and try to implement a logic that will only change the value of the field if it's
* a non-negative number
* a number value
so something like this could work
```
const changeMonthlyHoursInput = (value) => {
        if (Number(value) >= 0 || null) {
            setMonthlyhoursInput(value)
        }
    }
```
but for every text field that you think it's needed.
Then we shouldn't need to place this logic in here

src/styles/AllocationOverview.scss
```diff
@@ -0,0 +1,20 @@
...
+@media screen and (min-width: $sm) {
+    .AllocationOverview {
+        .edit-delete {
+            width: 70%;
+        }
+    }
+}
+
+@media screen and (max-width: $sm){
+    .AllocationOverview {
```
sofiarm21, line 14: I don't get why this media query is needed, you don't get the same result moving this to the `.AllocationOverview` at the top of the file? since `@media screen and (min-width: $sm)` is checking for bigger screens whatever you put outside the media query at `. AllocationOverview` should be applied in the cases where `max-width: $sm`. Pls check this and make sure it's not redundant code.

sofiarm21 (changes requested):

sofiarm21:
@otech47 what do you think about the payment tile edit button? 
In my opinion, I think a better approach was to make the tile hoverable changing the mouse cursor improving the UX, but placing a whole row with a new icon for each tile seems like too much space wasted, it's just my perception tho

src/styles/AllocationOverview.scss
Ovifer13: Ohhhh ok, I see I was doing it wrong! Going to fix it, ty!

src/components/AllocationOverview.js
Ovifer13: I'm going to fix those Issues with Issue #480. For now, I'm going to revert the changes of how It was before.

otech47:
> @otech47 what do you think about the payment tile edit button?
> In my opinion, I think a better approach was to make the tile hoverable changing the mouse cursor improving the UX, but placing a whole row with a new icon for each tile seems like too much space wasted, it's just my perception tho

Iterate on this idea:

Screen Shot 2021-07-28 at 3 47 32 PM

Ovifer13:
**Update on edit button**
[Payment Tiles with Edit button and hover transition](https://www.loom.com/share/1548ecaf936544868078c4e075233089)

otech47:
> **Update on edit button**
> [Payment Tiles with Edit button and hover transition](https://www.loom.com/share/1548ecaf936544868078c4e075233089)

@Ovifer13 nice but I think the animation is a bit too much if you can make it a lot more subtle. Specifically:

- the box shadow seems way too big it can be drastically reduced to be more subtle closer to how material UI makes hover effects
- the height of all the other tiles in the list also change in this animation so it just feels like way too many UI elements are moving around

Here is an article on UI transitions and what makes them good

https://uxplanet.org/functional-animation-in-ux-design-what-makes-a-good-transition-d6e7b4344e5e

Ovifer13:
> @Ovifer13 nice but I think the animation is a bit too much if you can make it a lot more subtle. Specifically:
> 
> * the box shadow seems way too big it can be drastically reduced to be more subtle closer to how material UI makes hover effects
> * the height of all the other tiles in the list also change in this animation so it just feels like way too many UI elements are moving around
> 
> Here is an article on UI transitions and what makes them good
> 
> https://uxplanet.org/functional-animation-in-ux-design-what-makes-a-good-transition-d6e7b4344e5e

**Update on animations**
@otech47 [Animations fix](https://www.loom.com/share/8a313ba10f4d4dd9b058e453ae14dd62)

otech47 (changes requested):
@Ovifer13 thanks definitely an improvement, can you do a couple more specific things here:

1. Make sure when hovering over the tile, the `cursor: pointer` CSS rule is set to change to the clickable icon
2. Reduce the size of the Edit icon and make sure the space between the icon and the top of the tile is the same as the icon and the right of the tile so it is symmetrical

Personally I think the fact that the text is moving around within the tile as the animation happens is still a bit jarring but the main objective has been accomplished so I don't mind merging as-is

Ovifer13:
**Update**

<a href="https://www.loom.com/share/9c5eaf338b6c46c5a349626b4f2f1194">
    <p>Trinary: Fix animation and edit button</p>
    
  </a>

otech47 (approved):

Issue #473, opened 2021-07-21
Make sure the header is always a fixed height


otech47:
When the title is too long to fit on 1 line the header should either make the text smaller and/or use truncated text to make sure the height does not allow for overflowing to a 2nd line 

![Screenshot_20210721-115436](https://user-images.githubusercontent.com/4914611/126520164-2feef3d7-4715-44eb-9cdc-4bacfeba2cc2.jpg)

PR #477, opened 2021-07-22
Header with fixed height 

Ovifer13:
**Issue #473**
The header was fixed so it won't overflow to a second line.

In the file `Navigation.js` the _h2_ tag was changed for a _Typography_ component because Typography uses a prop named noWrap that works as follows.
>If true, the text will not wrap, but instead will truncate with a text overflow ellipsis.
Note that text overflow can only happen with block or inline-block level elements (the element needs to have a width in order to overflow).

**The following image shows how the header will look in a Mobile Device**
image

### DIFF
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -5,8 +5,7 @@ import {
     AppBar,
     Box,
     Grid,
-    Typography,
-    makeStyles
+    Typography
 } from '@material-ui/core'
 import withWidth, { isWidthUp } from '@material-ui/core/withWidth'
 import { split } from 'lodash'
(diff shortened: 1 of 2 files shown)


src/components/Navigation.js
```diff
@@ -14,8 +16,20 @@ import { LOGO_URL, SMALL_LOGO_URL } from '../constants'
 import { capitalizeWord, matchTitlePage } from '../scripts/selectors'
 import { pageName } from '../reactivities/variables'
 
+const useStyles = makeStyles((theme) => ({
```
sofiarm21, line 19: There's a strong reason why you placed the styling up here and not in the `Navigation.scss` file? I don't see this pattern used anywhere else on the app

Ovifer13: The main reason was to make the header responsive. For some reason in the scss file, it wasn't working properly

sofiarm21: try moving this into the scss file using the `height` property and the `@media screen` to make it responsive, we should try to maintain styling in a separate file. @ at me on slack if you keep having problems with this

otech47 (approved):

Issue #480, opened 2021-07-23
Editing an Allocation allows letters and special characters on numeric fields


Ovifer13:
**Describe the bug**
In a project payment where you can see the allocations, if you choose to edit one, you can type letters and special characters inside fields that are supposed to only accept numbers. Only the total amount field on max budget won't let me type anything that is not a number.

**Screenshots**
**Prorated Monthly**
![image](https://user-images.githubusercontent.com/86666889/126813723-747a7a83-6aae-4dd6-9b80-6cd3b1345d5d.png)

**Max Budget**
![image](https://user-images.githubusercontent.com/86666889/126813856-b873e645-bfa7-424c-b992-4e57350cb8f9.png)

If you click on Edit it will save the incorrect inputs and the allocation will look as follows.
![image](https://user-images.githubusercontent.com/86666889/126814079-38bec312-a5a6-416d-90b9-0f4e634728e3.png)

otech47:
Yeah we should have validation functions that run on those fields every time they change so that only safe number values can be input there

otech47:
I know this can be done with redux-form but there may be other ways

https://redux-form.com/6.5.0/examples/material-ui/

And @sofiarm21 should know better where these fields are exactly but found one here:

https://github.com/setlife-network/trinary/blob/develop/src/components/RateMaxBudgetForm.js

PR #482, opened 2021-07-26
Allocations with wrong inputs

Ovifer13:
**Issue #480**

**Description:**

When creating an Allocation a user was able to input letters, negative numbers, and special characters when setting the hourly rate and expected monthly rate in prorated monthly and max budget options.

**Proof**
[Changes demostration](https://www.loom.com/share/d06fd60a2d9d4f4fbf2bf651bb3a0dbd)

### DIFF
--- a/src/scripts/validation.js
+++ b/src/scripts/validation.js
@@ -1,0 +1,7 @@
+export const validatePositiveNumbers = (newValue, preValue) => {
+    if (Number(newValue) >= 0 || null) {
+        return newValue
+    } else {
+        return preValue
+    }
+}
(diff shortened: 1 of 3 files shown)


otech47 (changes requested):
I think we should refactor this and a file `/src/scripts/validation.js` and put a separate `validatePositiveNumbers` function with this logic that is repeated 3 times.

Then we can import that function and call it in each of these cases

then the validation.js file can serve as a module where we can add more validation functions where a field input needs to meet certain requirements before being passed on

sofiarm21 (approved):

src/scripts/validation.js
```diff
@@ -0,0 +1,7 @@
+export const validatePositiveNumbers = (newValue, preValue) => {
```
otech47, line 1: NIT suggestion: `newValue, oldValue` or `nextValue, currentValue`

otech47 (approved):

Issue #424, opened 2021-06-04
Refactor AddProjectFromGithub to run separate queries for orgs+repos


otech47:
Rationale: https://www.loom.com/share/38dee712127349b39f835da740849cbc

Currently this feature fetches all the data needed in 1 API call

We should separate fetching the list of organizations, then when an organization is selected we fetch the repos for only that selected organization

![Screen Shot 2021-06-03 at 8 02 56 PM](https://user-images.githubusercontent.com/4914611/120726701-b4b04c00-c4a6-11eb-80fa-6ab4b85c568b.png)

otech47:
@mecoccaro

PR #483, opened 2021-07-27
Separated queries for orgs and repos

mecoccaro:
**Issue #424**
**Description**

We are separating the queries on the Add project from github view, so the selection of the organization has its own querie separated from the repo selection. Currently when you select an organization it makes the request and gets all the repos from that organization.

Implementation proof
https://www.loom.com/share/4d4774214b1248159022c286dd785fae

Thanks to @sofiarm21 for the help!

### DIFF
--- a/api/handlers/github.js
+++ b/api/handlers/github.js
@@ -58,25 +58,20 @@ const github = module.exports = (() => {
         const octokit = new Octokit({
             auth: params.auth_key
         })
-        let res
-        if (params.itsOrganization) {
-            console.log('org')
-            res = await octokit.repos.listForOrg({
+        const res = 
+        params.isOrganization
+            ? await octokit.repos.listForOrg({ 
                 org: params.organizationName,
                 per_page: 100,
                 sort: 'updated',
                 direction: 'desc',
             })
-        } else {
-            console.log('user')
-            res = await octokit.repos.listForUser({
+            : await octokit.repos.listForUser({
                 username: params.organizationName,
                 per_page: 100,
                 sort: 'updated',
                 direction: 'desc',
             })
-        }
-        
         if (res.status == 200) {
             return res.data
         } else {
--- a/api/handlers/github.js
+++ b/api/handlers/github.js
@@ -58,8 +58,8 @@ const github = module.exports = (() => {
         const octokit = new Octokit({
             auth: params.auth_key
         })
-        const res = await octokit.repos.listForOrg({
-            org: params.organizationName,
+        const res = await octokit.repos.listForUser({
+            username: params.organizationName,
             per_page: 100,
             sort: 'updated',
             direction: 'desc',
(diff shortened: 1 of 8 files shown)


api/handlers/github.js
```diff
@@ -58,12 +58,24 @@ const github = module.exports = (() => {
         const octokit = new Octokit({
             auth: params.auth_key
         })
-        const res = await octokit.repos.listForAuthenticatedUser({
-            affiliation: ['owner'],
-            per_page: 100,
-            sort: 'updated',
-            direction: 'desc',
-        })
+        let res
```
sofiarm21, line 61: @mecoccaro We want to use `const` whenever it's possible, so refactor this code to maybe something like `const res = {}` and then assign the values that you consider that are relevant to the object structure. I prefer having multiple `const` variables than just a few using `let` or `var`

@otech47 shared with me once an article about why we should enforce `const` over `let` and `var` but I couldn't find it, maybe he still has it pinned somewhere

sofiarm21 (changes requested):

api/handlers/github.js
otech47: https://codeburst.io/javascript-var-let-or-const-which-one-should-you-use-2fd521b050fa

api/modules/automations.js
```diff
@@ -33,31 +33,34 @@ const automations = module.exports = (() => {
...
+
+        if (organizations[0].name != params.organizationName) { // Check if its an org or user
+            itsOrganization = true
+        }
+
         const repos = await github.fetchRepos({
-            auth_key: params.auth_key
+            auth_key: params.auth_key,
+            organizationName: params.organizationName,
+            itsOrganization
```
otech47, line 46: I recommend we try to make the language of this logic more clear considering the following..

1. `isOrganization` and not `itsOrganization` because it may be confused with the possessive form "its" which means this organization may belong to some "it"
2. Remember to [minimize require knowledge](https://github.com/setlife-network/trinary/blob/develop/CONTRIBUTING.md#1-minimize-required-knowledge) you want to imagine that you know nothing about this logic before reading it, so as a reader I am asking why you are passing an `organizationName` if it's possible that `isOrganization` will be false? This can be related to @sofiarm21's idea about "Accounts" vs "Organizations" so it is more clear that it is possible for this name to either belong to a GitHub Organization versus a GitHub User

otech47 (changes requested):
See comments

api/handlers/github.js
mecoccaro: The thing here is that as the article says I know that the value will change, if I change it to const in this case, when we change the org it will cause an error because of the change of value on the constant.

otech47 (approved):
Cleanup to be addressed in https://github.com/setlife-network/trinary/issues/495

Issue #481, opened 2021-07-24
Improve Add/Edit Allocations UX for date selection


otech47:
As a user I want to be able to easily click from a set of option buttons that automatically calculate the start and end dates so that I don't have to click through so many dates/months/fields or count to figure out end dates 

As a user I want to more easily set a term length for a contributor to be allocated paid work on a project so that I don't have to think in dates but rather months/weeks/years from a certain start date 

The set of default option buttons might be:
- 1-4 weeks
- 1-12 months
- 1-5 years

Or maybe instead of/in addition to option buttons we can have a flexible and customizable way to pick start and end dates

Here is an example from AWS Cloudwatch that we can iterate on or even simplify

![Screenshot_20210724-130557](https://user-images.githubusercontent.com/4914611/126875943-68f3cceb-85c8-488f-ae67-aa666a71fbbe.jpg)
![Screenshot_20210724-130612](https://user-images.githubusercontent.com/4914611/126875944-81a04ab3-48b2-4a99-852d-4a5c6aa70766.jpg)

PR #484, opened 2021-07-28
Improve Add and Edit Allocations date selection

Ovifer13:
**Issue #481**
Sometimes selecting dates from a calendar can be difficult, so we are improving the way people can select weeks, months, and years for when they create or edit an allocation.


Ovifer13:
This functionality should add weeks, months, and years from the start date to the end date or add the same amount of time to both?
For example. The user sets a starting date and can add weeks, months, and years only to the end date with the new functionality?
image

@otech47 @sofiarm21 @mecoccaro

Ovifer13:
I already did the accordion but Instead of Quick Select what name can we use there?
image

Ovifer13:
**Update**
<a href="https://www.loom.com/share/e7a331c96a1145b9b353a2d5a2b347b0">
    <p>Trinary: Date Quick Select explanation - Watch Video</p>
    
  </a>

src/components/DateInput.js
```diff
@@ -0,0 +1,299 @@
...
+                                            variant={`${selectedWeek == 4 
+                                                ? 'contained' 
+                                                : 'outlined'
+                                            }`}
+                                            onClick={() => getWeekMonthYears(4)}
+                                        >
+                                            4
+                                        </Button>
+                                    </Grid>
+                                    <Grid item xs={2} />
```
sofiarm21, line 128: To make a more maintainable code do the following
```
const renderWeeks = () => {
   const weeks = [1,2,3,4]
   return weeks.map(w => {
      <Grid item xs={2}>
            <Button 
                    color='primary'
                     variant={`${selectedWeek == w 
                          ? 'contained' 
                           : 'outlined'
                        }`}
                         onClick={() => getWeekMonthYears(w)}
             >
                           {w}
             </Button>
           </Grid>
        <Grid item xs={2} />

   }
}
```

Now all you have to do is call the function `renderWeeks`, this way you are saving a lot of lines of code and if we want to change the weeks for any reason we only have to change the array. Do the same for months and years.

src/styles/DateInput.scss
```diff
@@ -0,0 +1,15 @@
+.DateInput {
+    .MuiPaper-elevation1 {
+        box-shadow: none;
+        border: 1px solid #F1F1F1;
+        color: #828282;
```
sofiarm21, line 5: Use the color variables. See `ContributorTile.scss` for reference

sofiarm21 (changes requested):

src/components/DateInput.js
Ovifer13: Thanks Sofia for this method, the code now looks cleaner!

sofiarm21 (approved):

Issue #373, opened 2021-03-16
UI Polish


otech47:
- [x] SetLife logo is misaligned with title of the header
- [x] Fix minor alignment issues on Home Page
![File (1)](https://user-images.githubusercontent.com/4914611/111328957-18b03a80-8645-11eb-8a9d-518cb977d93d.jpg)
![File](https://user-images.githubusercontent.com/4914611/111328959-1948d100-8645-11eb-9b77-416984b4f9bc.jpg)
- [x] Put client name under project name on ProjectTile to prevent bad overflow and left align all text
- [x] Change icon on ProjectTile
- [x] Iterate UI of project overview top panel. The blue background might be too much but maybe tweaking borders and font colors could improve this look
- [x] All buttons that update/save data should say Save instead of Edit but Edit is appropriate for bringing up an overlay to change data before saving
- [x] Fix alignment on mobile/small width view:
![File (2)](https://user-images.githubusercontent.com/4914611/111329799-c6bbe480-8645-11eb-96d9-8e1d93e3cae6.jpg)
- [x] Missing label on Project Start date
- [x] Group budget timeframe and budget amount so it makes more sense
- [x] Make it more obvious that you can edit payment allocations

PR #486, opened 2021-08-03
UI Polish

Ovifer13:
**Description Issue #373**
Different UI problems were fixed on this issue:

1. Logo misaligned with the header
image

2. Minors alignments on the homepage

3. In `ProjectTile` the **client name** now is under the **project name** and the icon changed
image

4. All buttons that update/save data now say **Save** instead of **Edit** 
5. **Budget timeframe** and **budget amount** now are grouped
image

Tasks that were not mentioned were already fixed.

### DIFF
--- a/src/styles/ProjectSummary.scss
+++ b/src/styles/ProjectSummary.scss
@@ -1,0 +1,12 @@
+.ProjectSummary{
+    .MuiBox-root-14{
+        border-radius: none;
+    }
+    border-radius: 1rem;
+}
+
+@media screen and (min-width: $sm) {
+    .ProjectSummary {
+
+    }
+}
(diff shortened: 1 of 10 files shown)


Ovifer13:
**Update**
In component `ProjectSummary` I didn't want to do big changes since this is going to get a refactor and it also looks good like this for now, so I ended changing the color of the edit button and adding some _border-radius_ to the overview
image

src/styles/ProjectSummary.scss
```diff
@@ -0,0 +1,12 @@
+.ProjectSummary{
+    .MuiBox-root-14{
```
otech47, line 2: Should there be space before the opening brace? @sofiarm21 not sure if we are being strict about this..

otech47 (approved):

Issue #489, opened 2021-08-05
Long project names overfill boxes 


mecoccaro:
On the contributor detail view, if we have a long project name it will overfill the box or it overlaps with the icon.
![imagen](https://user-images.githubusercontent.com/36824674/128428813-b07a97a8-4bbb-42bb-879b-5aa7df6de929.png)

PR #493, opened 2021-08-12
Long project names overfilling boxes

Ovifer13:
**Issue #489**
**Description**
Project names in `AllocationTile.js` and `ProjectTile.js` were overfilling from the boxes.

To fix this problem we can use the _overfill_ prop from the `<Box>` component of material-ui. This prop will make the text inside a Box to hide when it is longer than the box width.
[Display - Material-UI](https://material-ui.com/es/system/display/)

**Proof**
image


sofiarm21:
@Ovifer13 
Could you make an ellipsis effect for when the project name is too long it doesn't break into another line? so all the boxes would be the same height. This is on `/home/projects`
Screen Shot 2021-08-13 at 9 16 06 AM

otech47 (changes requested):
Agree with this: https://github.com/setlife-network/trinary/pull/493#issuecomment-898454649

Ovifer13:
> @Ovifer13
> Could you make an ellipsis effect for when the project name is too long it doesn't break into another line? so all the boxes would be the same height. This is on `/home/projects`
> Screen Shot 2021-08-13 at 9 16 06 AM

**Update**
image

sofiarm21 (approved):
@Ovifer13 notice the little change I made in last commit

Ovifer13:
> @Ovifer13 notice the little change I made in last commit

I see, I will do this every time I see code like that. Thank you, Sofia!

otech47 (approved):

Issue #446, opened 2021-07-07
Containerize the project with Docker


otech47:
Create a Dockerfile for frontend, backend, and/or both

sofiarm21:
@otech47 Never done this before but it seems interesting so I will assign myself and keep everyone updated on how It goes

otech47:
@sofiarm21 check the Dockerfile on Careerscore for an example

otech47:
- Combine the 2 Dockerfiles into 1 `/dockerfiles/Dockerfile-deploy`
- Create a new `Dockerfile-develop` (uses `npm run ui` instead of `npm run build` and also starts the web server with `node server`)

As a developer, I want to be be able to clone the repo from scratch and run a set of docker commands that set up the development environment without having to use `npm` commands. (Do we need `docker-compose` for this?)

As a system administrator, I want to be able to deploy the app to Heroku and see that the Dockerfile is used to build and deploy the application. (see https://devcenter.heroku.com/articles/build-docker-images-heroku-yml)

otech47:
@sofiarm21  not sure if you saw this already

https://engageinteractive.co.uk/blog/making-docker-faster-on-mac

PR #496, opened 2021-08-15
Docker files

sofiarm21:


otech47 (changes requested):
https://github.com/setlife-network/trinary/issues/446#issuecomment-902750757

Issue #487, opened 2021-08-05
Payments with no date_paid seem to be sorting incorrectly


otech47:
The payments with date_paid set are sorting correctly with the most recent at the top, but if dae_paid is not set then it is showing the oldest first instead of most recent (by date_incurred)

![Screen Shot 2021-08-05 at 9 25 20 AM](https://user-images.githubusercontent.com/4914611/128357824-e4ecf95a-22f7-4e07-a026-5ec6f9c8a659.png)

PR #497, opened 2021-08-18
Payments with no date paid sorting incorrectly

Ovifer13:
**Issue #487**
**Description**
Payments were being sorted by date_paid, this was causing that payments without date_paid were not being sorted.
To fix this problem the sort was changed from date_paid to date_incurred in `ClientPayments.js`

**Proof**
image

### DIFF
--- a/src/components/ClientPayments.js
+++ b/src/components/ClientPayments.js
@@ -18,7 +18,7 @@ const ClientPayments = (props) => {
     if (loading) return <LoadingProgress/>
     if (error) return `Error! ${error.message}`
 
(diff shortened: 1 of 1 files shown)


otech47:
@Ovifer13 note the commit here: https://github.com/setlife-network/trinary/pull/497/commits/00d0bdc4a535d47517fcb5a9dd834106ae8f1405

I was able to improve upon this by adding `date_incurred` as a 2nd backup sorting option since `orderBy` supports this

that was the most recent confirmed payments come up first, then at the bottom the unconfirmed payments will show next

otech47 (approved):

Issue #452, opened 2021-07-13
Only create Stripe Invoices if the currency is USD


otech47:
Currently when creating a Payment we check if the client has an `external_uuid` before trying to create a linked Stripe Invoice

We should also add another check on the `currency` as well

We do not care to link Trinary and Stripe unless the currency is USD because Stripe does not support BTC/SATS

We could consider supporting EUR or MEX but we would need to lock the currency field on the Client because Stripe will not allow different currencies to be used after an invoice has been created. If we decide to do this it will be a separate issue

sofiarm21:
In an effort of scoping this issue we will do the following:
1. This issue will only contain the implementation of this `We do not care to link Trinary and Stripe unless the currency is USD`
2. I'll proceed and create a new issue for this `lock the currency field on the Client because Stripe will not allow different currencies to be used after an invoice has been created` after giving it a further thought

otech47:
## From Stripe to Trinary

- When creating Invoices in Stripe, the connected Client in Trinary will have its currency changed to USD (the currency of the invoice) and then will be locked from updating
- Do not create Trinary Payments if there are already payments in a different currency (should we "disconnect" the Client from the Customer in this case and delete the `external_uuid`)

## From Trinary to Stripe

- When creating a Trinary Payment, it should only create a Stripe invoice if the currency is supported (USD, EUR, MXN) only after updating the currency of the Stripe Customer which will then be locked after the invoice is created

PR #499, opened 2021-08-18
Only create invoices when the currency is USD

mecoccaro:
**Issue #452**
**Description**

The task for this issue are:

## From Stripe to Trinary

- When creating Invoices in Stripe, the connected Client in Trinary will have its currency changed to USD (the currency of the invoice) and then will be locked from updating
- Do not create Trinary Payments if there are already payments in a different currency (should we "disconnect" the Client from the Customer in this case and delete the `external_uuid`)

## From Trinary to Stripe

- When creating a Trinary Payment, it should only create a Stripe invoice if the currency is supported (USD, EUR, MXN) only after updating the currency of the Stripe Customer which will then be locked after the invoice is created

### DIFF
--- a/api/modules/budgeting.js
+++ b/api/modules/budgeting.js
@@ -4,6 +4,33 @@ const budgeting = module.exports = (() => {
     const db = require('../models')
     const clientManagement = require('./clientManagement')
 
+    const checkExistentPaymentsWithExternalId = async (params) => {
+        const existentPayments = await db.models.Payment.findAll({
+            where: {
+                external_uuid: params.id
+            }
+        })
+
+        if (existentPayments) {
+            for (const payment of existentPayments) {
+                const client = db.models.Client.findOne({
+                    where: {
+                        id: payment.client_id
+                    }
+                })
+
+                if (client.currency != 'USD') {
+                    try {
+                        const deletedCustomer = clientManagement.deleteClientUuid({ id: client.external_uuid })
+                        console.log(`stripe customer disconnected ${deletedCustomer}`)
+                    } catch (err) {
+                        console.log(`An error ocurred: ${err}`)
+                    }
+                }
+            }
+        }
+    }
+
     const createPayment = async ({ paymentInformation }) => {
         const client = await clientManagement.findClientWithExternalId({
             id: paymentInformation.customer_id
@@ -16,6 +43,8 @@ const budgeting = module.exports = (() => {
             date_paid,
             external_uuid_type
         } = paymentInformation
+
+        const checkExistentPayments = await checkExistentPaymentsWithExternalId({ id: external_uuid })
 
         if (client) {
             return db.models.Payment.create({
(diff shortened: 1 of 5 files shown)


sofiarm21 (approved):

sofiarm21:
@mecoccaro pls resolve the file conflicts and remember to record the implementation proof again

api/modules/budgeting.js
```diff
@@ -1,9 +1,33 @@
...
+    const checkExistingPaymentsWithExternalId = async (params) => {
+        const existingPayments = await db.models.Payment.findAll({
+            where: {
+                external_uuid: params.id
+            }
+        })
+        
+        if (existingPayments) {
+            existingPayments.forEach(async payment => {
+                const client = await clientManagement.findClientWithId(payment.client_id)
```
otech47, line 17: Not needed since `client` can be passed in as a parameter

api/modules/budgeting.js
```diff
@@ -1,9 +1,33 @@
 const moment = require('moment')
+const { DEFAULT_STRIPE_CURRENCY } = require('../config/constants')
 
 const budgeting = module.exports = (() => {
     const db = require('../models')
     const clientManagement = require('./clientManagement')
 
+    const checkExistingPaymentsWithExternalId = async (params) => {
```
otech47, line 8: - Change to `checkForMismatchedClientPayments`
- Use `client_id` on the Payments to match `client.id` passed in from the function call

api/modules/budgeting.js
```diff
@@ -1,9 +1,33 @@
...
+        const existingPayments = await db.models.Payment.findAll({
+            where: {
+                external_uuid: params.id
+            }
+        })
```
otech47, line 13: Refactor to a smaller function called `getPaymentsWithClientId`

api/modules/budgeting.js
```diff
@@ -1,9 +1,33 @@
...
+        
+        if (existingPayments) {
+            existingPayments.forEach(async payment => {
+                const client = await clientManagement.findClientWithId(payment.client_id)
+
+                if (client.currency != DEFAULT_STRIPE_CURRENCY) {
```
otech47, line 19: This comparison should be between `client.currency` and `stripeInvoice.currency`

api/modules/budgeting.js
```diff
@@ -1,9 +1,33 @@
...
+                    try {
+                        const deletedCustomer = await clientManagement.deleteClientUuid({ id: client.external_uuid })
```
otech47, line 21: Change to `disconnectedCustomer` to remove the confusion of the client being "deleted" when it is actually being updated to clear out the value of the `external_uuid`

api/modules/budgeting.js
```diff
@@ -65,6 +89,18 @@ const budgeting = module.exports = (() => {
                 external_uuid_type: 'STRIPE'
             }
         })
+        const client = await clientManagement.findClientWithExternalId({
+            id: stripeInvoice.customer
+        })
+
+        try {
+            await checkExistingPaymentsWithExternalId({ id: stripeInvoice.id })
```
otech47, line 97: Store the return value as a variable

```
const paymentsDoNotMatch = await checkForMismatchedClientPayments(...)
```

OR 

use a `throw` in the `checkForMismatchedClientPayments` function

api/modules/budgeting.js
```diff
@@ -65,6 +89,18 @@ const budgeting = module.exports = (() => {
...
+            client.currency = DEFAULT_STRIPE_CURRENCY //Stripe invoices will always have USD as currency
```
otech47, line 98: This should be changing the client currency to match `stripeInvoice.currency` instead of `DEFAULT_STRIPE_CURRENCY`

otech47 (changes requested):

Issue #475, opened 2021-07-21
Add a Search Contributors to the ProjectContributors page


otech47:

Ovifer13:
I can try this one but I will need help with the backend

Ovifer13:
[Material UI Autocomplete](https://material-ui.com/es/components/autocomplete/)

PR #500, opened 2021-08-19
Search contributor component

Ovifer13:
**Issue #475**
**Description**
A search bar to filter contributors on `ProjectContributors` page

### DIFF
--- a/src/components/ProjectContributors.js
+++ b/src/components/ProjectContributors.js
@@ -4,8 +4,10 @@ import {
     Box,
     Grid,
     Typography,
-    Link
+    Link,
+    TextField
 } from '@material-ui/core/'
+import Autocomplete from '@material-ui/lab/Autocomplete'
 import {
     differenceBy,
     filter
@@ -33,6 +35,7 @@ const ProjectContributors = (props) => {
     const [githubContributors, setGithubContributors] = useState([])
     const [openAddAllocationDialog, setOpenAddAllocationDialog] = useState(false)
     const [contributorClicked, setContributorClicked] = useState(null)
+    const [autoComplete, setAutoComplete] = useState()
     const handleAddAllocationClose = (value) => {
         setOpenAddAllocationDialog(false)
     }
@@ -151,6 +154,12 @@ const ProjectContributors = (props) => {
         })
     }
 
+    const searchContributosToAdd = []
+
+    contributorsToAdd.map(c => {
+        searchContributosToAdd.push(c.name)
+    })
+
     return (
         <Grid container className='ProjectContributors'>
             <h1>
@@ -217,6 +226,15 @@ const ProjectContributors = (props) => {
                 ? isNotAContributor()
                 : null
             }
+            <Grid item xs={12}>
+                <Autocomplete
+                    options={searchContributosToAdd}
+                    renderInput={(params) => <TextField {...params} label='Search contributor' variant='outlined' />}
+                    autoComplete
+                    clearOnEscape
+                    onChange={(event, value) => setAutoComplete(value)}
+                />
+            </Grid>
             <Grid item xs={12}>
                 <Box>
                     <Grid container>


otech47:
The lag is due to too many contributors being rendered (be sure to test with a lower number of array elements to confirm that this is the root cause) so we can:

1. Explore optimizations involving lazy loading React components
2. Fallback solutions might be to add pagination and a Load More button

Ovifer13:
@otech47 after researching and trying different things I couldn't solve the problem of the lag but now It looks a bit faster with all the changes I did.

src/components/ContributorList.js
```diff
@@ -0,0 +1,188 @@
...
+                                ? <Suspense fallback={<div>Loading...</div>}>
+                                    <ActiveContributors
+                                        active
+                                        contributors={activeContributors}
+                                        project={project}
+                                        addAllocation={addAllocation}
+                                    />
+                                </Suspense>
+                                : <ContributorsEmptyState active/>
+                        }
```
sofiarm21, line 123: Fix styling

src/components/ContributorList.js
```diff
@@ -0,0 +1,188 @@
...
+                    </Grid>
+                </Box>
+                <Box my={[2, 5]}>
+                    <Typography align='left' variant='h5'>
+                        {`Upcoming contributors`}
+                    </Typography>
+                    <Grid container>
+                        {
+                            upcomingContributors.length != 0
```
sofiarm21, line 132: Fix styling

src/components/ContributorList.js
```diff
@@ -0,0 +1,188 @@
...
+                            contributorsToAdd.length != 0
+                                ? <Suspense fallback={<div>Loading...</div>}>
+                                    <Contributors
+                                        contributors={contributorsToAdd}
+                                        project={project}
+                                        addAllocation={addAllocation}
+                                    />
+                                </Suspense>
+                                : <ContributorsEmptyState/>
+                        }
```
sofiarm21, line 168: Fix styling

src/components/ContributorList.js
```diff
@@ -0,0 +1,188 @@
...
+                    </Grid>
+                </Box>
+                <Box my={[2, 5]} py={5}/>
+            </Grid>
+            <Grid item xs={12}>
+                {
+                    contributorClicked &&
```
sofiarm21, line 175: Fix styling

src/components/ProjectContributors.js
```diff
@@ -129,122 +130,15 @@ const ProjectContributors = (props) => {
...
-                }
-            </Grid>
-        </Grid>
+        <ContributorList 
+            contributors={contributors}
+            project={project}
+            activeContributorsAllocated={activeContributorsAllocated}
+            upcomingContributorsAllocatedOnly={upcomingContributorsAllocatedOnly}
+            errorGithubContributors={errorGithubContributors}
+            isNotAContributor={isNotAContributor}
```
sofiarm21, line 140: @Ovifer13 Instead of passing this from the parent, we can just import the component `GithubAccessBlocked` directly in the `ContributorList` component and with a conditional render the component itself, I don't remember why there was a function to return the component but I think that could be changed

src/components/UpcomingContributors.js
```diff
@@ -0,0 +1,31 @@
...
+        contributors,
+        active,
+        project,
+        addAllocation
+    } = props
+
+    return (
+        contributors.map(c => {
+            return (
+                <Grid item xs={12} md={6} key={c.id}>
```
sofiarm21, line 18: Add the className 'UpcomingContributors' in here even if it doesn't have a sass file, this helps when debugging directly in the web inspector. As a personal preference, I like to return an element instead of a map function, so I would create a return with the item row an inside of it call a renderUpcomingContributors function with the map

src/components/ActiveContributors.js
```diff
@@ -0,0 +1,31 @@
...
+        contributors,
+        active,
+        project,
+        addAllocation
+    } = props
+
+    return (
+        contributors.map(c => {
+            return (
+                <Grid item xs={12} md={6} key={c.id}>
```
sofiarm21, line 18: Same comment from UpcomingContributors

sofiarm21 (changes requested):

sofiarm21:
@Ovifer13 Great pr, go ahead and create an issue for improving load time and performance and another one for changing the x color in the text field to setlife primary and make it clickable on hover

src/components/ActiveContributors.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import {
+    Grid
+} from '@material-ui/core'
+import ContributorTile from './ContributorTile'
+
+const ActiveContributors = React.memo((props) => {
```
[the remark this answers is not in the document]
Ovifer13: Yeah, I felt the difference between using and not using memo in performance, but it's still lagging.

src/components/ContributorList.js
```diff
@@ -0,0 +1,188 @@
...
+        isNotAContributor
+    } = props
+
+    const [openAddAllocationDialog, setOpenAddAllocationDialog] = useState(false)
+    const [contributorClicked, setContributorClicked] = useState(null)
+    const [search, setSearch] = useState('')
+
+    const ActiveContributors = React.lazy(() => import('./ActiveContributors'))
+    const UpcomingContributors = React.lazy(() => import('./UpcomingContributors'))
+    const Contributors = React.lazy(() => import('./Contributors'))
```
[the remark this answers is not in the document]
Ovifer13: This type of import is used to make components do lazy loading, memo is different from this.

src/components/ContributorList.js
sofiarm21: @Ovifer13 
```
{contributorsToAdd.length != 0
    ? (
        <Suspense 
            fallback={
                <div>Loading...</div>
            }
        >
            <Contributors
                contributors={contributorsToAdd}
                project={project}
                addAllocation={addAllocation}
            />
        </Suspense>
    ) : <ContributorsEmptyState/>
}
```

src/components/ContributorList.js
sofiarm21: @Ovifer13 
```
{contributorClicked &&
```

src/components/ContributorList.js
sofiarm21: @Ovifer13 
```
{upcomingContributors.length != 0
```

sofiarm21 (approved):

sofiarm21 (approved):

Issue #429, opened 2021-06-11
Add pagination to repository selection in the AddProjectFromGithub when a org has > 100 repos


otech47:
related #424

- Determine if there are more repos to fetch by counting the returned repos > 99 and show a Load More button
- Clicking the Load More button will fetch again with a `page` parameter = 2 which increments each time a new set of 100 is returned, then check again to see if < 99 are returned to hide the Load More button

https://octokit.github.io/rest.js/v18#repos

Be sure to reset the page counter whenever the organization is checked and employ proper state management

To test, change the `per_page` parameter to something small

PR #503, opened 2021-08-27
Add pagination to the repository selection

mecoccaro:
**Issue #429 **
**Description**

We want to add a pagination to repository selection in the AddProjectFromGithub when a org has > 100 repos. Using the page property of the Github API we can have a pagination if the org/user has more than 100 repos.

**Implementation Proof**
https://www.loom.com/share/5fe4f9b23964407da88c44e05099734f


sofiarm21 (approved):
@mecoccaro This looks great, we can create another issue that every time the user clicks load more the selected project doesn't change

otech47 (approved):

Issue #414, opened 2021-05-21
Improve syncTogglProject mutation function


otech47:
https://www.loom.com/share/65e9919c82464597a835f0d4983f15eb

otech47:
- Timestamp that gets added to the database was not properly formatted

- Add ability to extract `toggl_id` from the `toggl_url`
- 1. You can make sure the `toggl_id` gets updated when the Project is updated with the Toggl URL form the URL
- 2. Add an extra logic check in `syncTogglProject` that checks for a Toggl URL before failing

- Simplify the parameter logic to take a `toggl_url` which should have the toggl_id contained in it

PR #523, opened 2021-09-15
Improve Toggl functionality

mecoccaro:
**Issue #414**
**Description**

Improve the Toggl sync functionality for syncing the time entries per collaborator, toggl ID and workspace ID.

**Taks**
- Timestamp that gets added to the database was not properly formatted

- Add ability to extract `toggl_id` from the `toggl_url`
- 1. You can make sure the `toggl_id` gets updated when the Project is updated with the Toggl URL form the URL
- 2. Add an extra logic check in `syncTogglProject` that checks for a Toggl URL before failing

- Simplify the parameter logic to take a `toggl_url` which should have the toggl_id contained in it

**Currently testing**

### DIFF
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -29,12 +29,16 @@ const automations = module.exports = (() => {
 
     const getTogglPropertiesFromURL = async (params) => {
         // Url example https://track.toggl.com/3070292/projects/156669532
-        URL = params.toggl_url
+        URL = params
         const splitedUrl = URL.split(/\//)
-        const workspaceId = splitedUrl[2]
-        const projectId = splitedUrl[4]
+        try {
+            const workspaceId = splitedUrl[3]
+            const togglId = splitedUrl[5]
 
-        return { workspaceId, projectId }
+            return { workspaceId, togglId }
+        } catch (err) {
+            console.log('an error ocurred while spliting the toggl url: ', err)
+        }
     }
 
     const getUserOrganizations = async (params) => {
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -3,6 +3,40 @@ const db = require('../models')
 
 const automations = module.exports = (() => {
 
+    const getOrganizationRepos = async (params) => {
+        const organizations = await getUserOrganizations({
+            auth_key: params.auth_key
+        })
+        const isOrganization = organizations[0].name != params.organizationName
+
+        const repos = await github.fetchRepos({
+            auth_key: params.auth_key,
+            organizationName: params.organizationName,
+            isOrganization
+        })
+        const organizationRepos = []
+        repos.map(r => {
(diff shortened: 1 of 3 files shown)


api/scripts/timeLogging.js
```diff
@@ -3,11 +3,12 @@ const db = require('../models')
 module.exports = (() => {
 
     const matchTimeEntry = async (timeEntry) => {
-        return db.models.TimeEntry.findOne({
+        const match = await db.models.TimeEntry.findOne({
             where: {
                 toggl_id: timeEntry.id
             }
         })
+        return match
```
sofiarm21, line 11: @otech47 are these not the same?
`
return db.models.TimeEntry.findOne({...})
`
and
```
const match = await db.models.TimeEntry.findOne({...})
return match
```
Based on this https://jakearchibald.com/2017/await-vs-return-vs-return-await/ should be the same, but I was making some tests on this sandbox https://codesandbox.io/s/priceless-andras-4de81?file=/src/index.js and I'm not getting the expected result

sofiarm21 (changes requested):

sofiarm21:
@mecoccaro please add an implementation proof

api/scripts/timeLogging.js
mecoccaro: As far as I know they are the same, I changed just to check the returned value con a log before the return statement. I will rollback that change. @sofiarm21

api/modules/automations.js
```diff
@@ -30,19 +30,15 @@ const automations = module.exports = (() => {
...
             let workspaceId = 0
-            splitedUrl.forEach(function(element, index) {
-                if (index == 5) {
-                    togglId = element
-                } else if (index == 3) {
-                    workspaceId = element
-                }
+            splitedUrl.map((element, index) => {
+                togglId = splitedUrl[5] ? splitedUrl[5] : 0
+                workspaceId = splitedUrl[3] ? splitedUrl[3] : 0
```
sofiarm21, line 40: @mecoccaro why this needs to be inside a map function?

api/scripts/timeLogging.js
```diff
@@ -19,21 +19,29 @@ module.exports = (() => {
     }
 
     const addTimeEntries = (params) => {
-        return Promise.all(params.timeEntries.map(async t => {
-            if (!(await matchTimeEntry(t))) {
-                const contributor = await matchContributor(t)
+        let contributor
+        return params.timeEntries.map(async timeEntry => {
+            if (!(await matchTimeEntry(timeEntry))) {
+                contributor = await matchContributor(timeEntry)
```
sofiarm21, line 25: delete `let contributor` line
and on line 25 do `const contributor = await matchContributor(timeEntry)`

sofiarm21 (changes requested):

Issue #495, opened 2021-08-13
Refactor GitHub account-repo fetching to use IDs over names


otech47:
https://github.com/setlife-network/trinary/pull/483/files#r687076768

- [x] Changing the `getGithubRepos` query to accept an ID
- [x] Refactor functions called in the resolvers/modules/handlers and[ introduce `accountType` into the user vs organization logic](https://github.com/setlife-network/trinary/pull/483#discussion_r687084252)
- [x] Pass ID instead of name on the frontend

PR #530, opened 2021-09-29
Refactor GitHub account repo to use an AccountId instead of name

mecoccaro:
**Issue #495**
**Description**

We wanted to use an Id instead of an username or organization name for the getGithubRepos query, so we changed the query so it can accept an accountId and then it uses a logic to check if its an user or an organization.

- [x] Changing the `getGithubRepos` query to accept an ID
- [x] Refactor functions called in the resolvers/modules/handlers and[ introduce `accountType` into the user vs organization logic](https://github.com/setlife-network/trinary/pull/483#discussion_r687084252)
- [x] Pass ID instead of name on the frontend

**impletentation proof**
https://www.loom.com/share/0d9f048aa01a41deaedd6b43f4005a42

### DIFF
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -65,7 +65,7 @@ const automations = module.exports = (() => {
         })
         const organizationRepos = []
         repos.map(r => {
-            if (r.owner.login == accountName) {
+            if (r.owner.login == githubAccount.handle) {
                 organizationRepos.push({
                     id: r.id,
                     name: r.name,
--- a/api/modules/automations.js
+++ b/api/modules/automations.js
@@ -30,20 +30,43 @@ const automations = module.exports = (() => {
     }
 
     const getOrganizationRepos = async (params) => {
-        const organizations = await getUserOrganizations({
-            auth_key: params.auth_key
-        })
-        const isOrganization = organizations[0].name != params.organizationName
         
+        const contributor = await db.models.Contributor.findOne({
+            where: {
+                github_id: params.accountId 
+            }
+        })
+
+        let organizations
+        let accountName
+        let isOrganization = false
+        if (contributor) {
+            const contributorSplit = contributor.github_handle.split(/\//)
+            accountName = contributorSplit[3]
+        } else {
+            organizations = await getUserOrganizations({
+                auth_key: params.auth_key
+            })
+        }
+
+        if (organizations) {
+            isOrganization = true
+            organizations.map((org, i) => {
+                if (org.id == params.accountId) {
+                    accountName = org.name
+                }
+            })
+        }
+        console.log(accountName, isOrganization)
         const repos = await github.fetchRepos({
             auth_key: params.auth_key,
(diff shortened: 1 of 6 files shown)


api/modules/automations.js
```diff
@@ -30,20 +30,43 @@ const automations = module.exports = (() => {
...
+
+        if (organizations) {
+            isOrganization = true
+            organizations.map((org, i) => {
+                if (org.id == params.accountId) {
+                    accountName = org.name
+                }
+            })
+        }
+        console.log(accountName, isOrganization)
```
sofiarm21, line 60: Always try to think about solutions without the use o `var` and `let` and use `const` instead
To achieve that in this logic you can do
```
const organizations = []
const githubAccount = {}
if (contributor) {
    githubAccount.handler = contributor.github_handle.split(/\//)[3]
 } else {
     organizations.push(
            ...await getUserOrganizations({
                          auth_key: params.auth_key
             })
       )
 }

if (organizations.length) {
            organizations.map((org, i) => {
                if (org.id == params.accountId) {
                    githubAccount.handle = org.name
                }
            })
}


const repos = await github.fetchRepos({
            auth_key: params.auth_key,
            organizationName: params.organizationName,
            accountName: githubAccount.handle,
            githubPageNumber: params.githubPageNumber,
            isOrganization: organizations.length
})
```

sofiarm21, line 60: Delete log

api/modules/automations.js
```diff
@@ -30,20 +30,43 @@ const automations = module.exports = (() => {
...
+            accountName = contributorSplit[3]
+        } else {
+            organizations = await getUserOrganizations({
+                auth_key: params.auth_key
+            })
+        }
+
+        if (organizations) {
+            isOrganization = true
+            organizations.map((org, i) => {
```
sofiarm21, line 54: On the maps there's no need to create the `i` variable if it's not going to be used

api/schema/resolvers/ContributorResolver.js
```diff
@@ -98,7 +98,7 @@ module.exports = {
             })
             return contributorOrganizations
         },
-        getGithubRepos: async (root, { organizationName, githubPageNumber }, { cookies, models }) => {
+        getGithubRepos: async (root, { githubPageNumber, accountId }, { cookies, models }) => {
```
sofiarm21, line 101: Instead of using `accountId` it is possible to use the `contributor_id`? (the pk value of the Contributors table)

sofiarm21 (changes requested):

api/schema/resolvers/ContributorResolver.js
mecoccaro: Right now the logic is using the accountId in order to check if the user selected an organization or an user from the list, if we use the contributor id we would need to check twice and the accountId would still be necessary to see if its an org or an user

otech47 (approved):

Issue #485, opened 2021-07-30
Lock the currency of the Client if Stripe-connected Payments exist


otech47:
In an effort of scoping this issue we will do the following:
1. This issue will only contain the implementation of this `We do not care to link Trinary and Stripe unless the currency is USD`
2. I'll proceed and create a new issue for this `lock the currency field on the Client because Stripe will not allow different currencies to be used after an invoice has been created` after giving it a further thought

_Originally posted by @sofiarm21 in https://github.com/setlife-network/trinary/issues/452#issuecomment-888470824_

otech47:
Be sure to do this as soon as a Stripe invoice is created (different webhook) to be as closely in-sync as possible

otech47:
When "disconnecting" the Client from the Customer in #499, we should also unlock the currency if it is locked

PR #539, opened 2021-10-08
Lock client currency

mecoccaro:
**Issue #485**
**Description**

We want to lock the currency of a client when a payment or invoice is created from stripe if the client is connected to the customer.
We check the external id of the client and the existing payments so we can lock the currency.
If the client is disconnected from the stripe customer the currency will be unlocked.

**Implementation proof**
https://www.loom.com/share/40e2faab117c40548e22f6a78b8e61c8

Second proof 
https://www.loom.com/share/43d7db34b66f446da90f026318e383be

### DIFF
--- a/src/components/ClientEditDialog.js
+++ b/src/components/ClientEditDialog.js
@@ -28,6 +28,7 @@ const ClientEditDialog = (props) => {
 
     const [updateClient, { data, loading, error }] = useMutation(UPDATE_CLIENT)
 
+    const [clientCurrencyLock, setClientCurrencyLock] = useState(false)
     const [clientName, setClientName] = useState(client.name)
     const [clientEmail, setClientEmail] = useState(client.email)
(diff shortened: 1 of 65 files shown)


src/components/ClientEditDialog.js
```diff
@@ -78,6 +79,10 @@ const ClientEditDialog = (props) => {
         } else {
             setDisableEdit(false)
         }
+
+        if (client.currencyLocked) {
+            setClientCurrencyLock(true)
+        }
```
sofiaromorales, line 85: Why do you need to add that attribute to a piece of state? I don't think that value is going to change over time with a user interaction so I would delete the `clientCurrecyLock and setClientCurrencyLock` and on the render function I would just type
`disabled={client.currencyLocked}`

sofiaromorales (changes requested):

otech47 (approved):

Issue #522, opened 2021-09-13
Change the x color inside the Search component in ProjectContributors component


Ovifer13:
**Description**
The color of the _x_ inside the search bar component needs to be changed to the SetLife primary color

**Proof**
![image](https://user-images.githubusercontent.com/86666889/133113050-63ddff78-effb-4a8a-9eb8-d765b8d1cd5a.png)

PR #542, opened 2021-10-11
Change color of the clear icon inside Search component in ProjectContributors

Ovifer13:
**Issue #522**
**Description**
The color from the clear icon was a color that was not from the palette, with this change the icon is different and the color is the SetLife blue.

**Proof**
image

**What color do you think that looks better for the clear Icon? Blue, black, or grey?**

**Should we change the clear icon to any of these options?**
image

**Should we add the search icon at the start of the search bar?**
image


sofiaromorales:
I do think that adding the search icon is a good idea, in light grey and when the users clicks on the field hide it

Issue #506, opened 2021-09-01
Reloading ProjectContributors on a project with limited GitHub permissions breaks the app


otech47:
https://www.loom.com/share/7061dc0f1dca447c9e125eafdafe2764

PR #545, opened 2021-10-15
Reloading contributors page bug

mecoccaro:
**Issue #506**
**Description**

When reloading de contributord page of a project where you doen't have all the permissions, the page crashed.
We fixed modifying the query for getting the project url to use a lazy query.

**Implementation proof**
https://www.loom.com/share/abc9819c43e246be87de25caad977c55

### DIFF
--- a/src/components/GithubAccessBlocked.js
+++ b/src/components/GithubAccessBlocked.js
@@ -1,28 +1,35 @@
-import React from 'react'
+import React, {useEffect, useState} from 'react'
 import {
     Box,
     Grid,
     Typography,
     Link
 } from '@material-ui/core'
-import { useQuery } from '@apollo/client';
+import {useLazyQuery, useQuery} from '@apollo/client';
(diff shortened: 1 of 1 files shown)


src/components/GithubAccessBlocked.js
```diff
@@ -1,28 +1,35 @@
...
+    const [getProject, {
         data: dataProject,
         loading: loadingProject,
         error: errorProject,
-    } = useQuery(GET_PROJECT, {
-        variables: {
-            id: Number(props.projectId)
+    }] = useLazyQuery(GET_PROJECT, {
+        onCompleted: dataProject => {
+            setProject(dataProject.getProjectById.github_url)
```
sofiaromorales, line 23: Two things can be done here,

1. set the `project` state value to the whole `dataProject.getProjectById` object and on the Link href set to`project.github_url`
2. change the state variable name to `projectUrl`. Is a little confusing having a variable called project that only stores the project url and not the complete project object

otech47 (approved):

PR #551, opened 2021-10-22
Build an script that imports all invoicely csv data to Stripe

mecoccaro:
**Issue #391**
**Description**

We build an scripts that take the Invoicely csv file from S3 bucket and import that data to Stripe.
We import the client to an stripe customer and create an invoice with each invoicely invoice.

**Implementation proof**
https://www.loom.com/share/fe31b2e6fd7a4f70877d9cd564dd6f2c


api/modules/dataSyncs.js
```diff
@@ -12,6 +13,44 @@ const { GITHUB, TOGGL } = require('../config/credentials')
...
+                    customerInformation = customer
+                } else if (!customersFromCsv.includes(csvData.Client)) {
+                    customerInformation = await stripe.createCustomer({ name: csvData.Client, email: null })
+                    customersFromCsv.push(csvData.Client)
+                }
+                await stripe.createInvoice({
+                    amount: parseInt(csvData.Total),
+                    external_uuid: customerInformation.id,
+                    currency: csvData.Currency
+                })
```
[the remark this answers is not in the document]
mecoccaro: The only way to set a date is using the due date and if we use this parameter it will create the invoice as a pending one, so when the date is already the invoice is charged, the invoices with an older date Stripe just ignore them.

api/modules/dataSyncs.js
```diff
@@ -12,6 +13,45 @@ const { GITHUB, TOGGL } = require('../config/credentials')
...
+        const invoiceFile = INVOICELY_CSV_PATH
+        const csvFile =
+            await amazon.fetchFile({ file: invoiceFile })
+                .then(file => {
+                    return invoicelyCodebase.modelCSV(file)
+                })
+                .catch(err => {
+                    console.log('error', err);
+                    return err.message
+                })
```
otech47, line 26: There is no need to use `await` together with `.then/catch` you can just do:

```
const csvFile = await amazon.fetchFile({ file: invoiceFile })
const modeledCsv = await invoicelyCodebase.modelCSV(csvFile)
```

If you need to catch an error use a try/catch or move this into the try/catch you already wrote below

```
const stripeCustomers = await stripe.listAllCustomers()
const customersFromCsv = []
try {
    const csvFile = await amazon.fetchFile({ file: invoiceFile })
    const modeledCsv = await invoicelyCodebase.modelCSV(csvFile)
    modeledCsv.map(async csvData => {
        ...
} catch (err) { }
```

api/modules/dataSyncs.js
```diff
@@ -12,6 +13,45 @@ const { GITHUB, TOGGL } = require('../config/credentials')
...
+                let customerInformation
+                const customer = stripeCustomers.data.find(customerData => customerData.name == csvData.Client)
+                csvData.Total = csvData.Total.replace(/,/g, '')
+                const amount = parseInt(csvData.Total) * 100
+                if (customer) {
+                    customerInformation = customer
+                } else if (!customersFromCsv.includes(csvData.Client)) {
+                    customersFromCsv.push(csvData.Client)
+                    customerInformation = await stripe.createCustomer({ name: csvData.Client, email: null })
+
```
otech47, line 40: Remove empty line

api/modules/dataSyncs.js
```diff
@@ -12,6 +13,45 @@ const { GITHUB, TOGGL } = require('../config/credentials')
...
+                    return err.message
+                })
+        const stripeCustomers = await stripe.listAllCustomers()
+        const customersFromCsv = []
+        try {
+            csvFile.map(async csvData => {
+                let customerInformation
+                const customer = stripeCustomers.data.find(customerData => customerData.name == csvData.Client)
+                csvData.Total = csvData.Total.replace(/,/g, '')
+                const amount = parseInt(csvData.Total) * 100
```
otech47, line 34: Probably would be good to add a `const amount = Number((parseInt(csvData.Total) * 100).toFixed(0))` to avoid floating point errors

Also might as well use `Number` over `parseInt`

```
const amount = Number((Number(csvData.Total) * 100).toFixed(0))
```

otech47 (changes requested):

Issue #554, opened 2021-10-25
Improve PaymentTile on ClientPage to include allocation functionality like tiles in ProjectPayments


otech47:
https://www.loom.com/share/c3a4f37c03db4aceb071ce4863573816?from_recorder=1&focus_title=1

- Retain the ability to edit the Payment
- Show allocations and include the project for each
- Allow adding new allocations to the payment (project must be selected)

PR #557, opened 2021-11-03
Improve PaymentTile on ClientPage

Ovifer13:
**Issue #554**
**Description**
In `ClientPage` the user should be able to allocate a contributor from the `PaymentTile`.

**Proof**
image

**Feedback**
1) Where should I place the Allocate button?
2) What do you think of moving the Date next to the amount?
 Right now it looks like this 
image

 How It would look if we move the Date next to the amount
image


Ovifer13:
**Update**
image

sofiaromorales (approved):

otech47 (approved):

Issue #546, opened 2021-10-15
Use finalized_at timestamp for date_incurred on Invoice webhook


otech47:
When a Stripe invoice is finalized and set to `ready_to_allocate` to be triggered by the webhook to create a Payment, the logic correctly brings in the `date_incurred` data when creating or updating the Trinary Payment. If provided in the metadata, it uses the override value or defaults to the Invoice creation timestamp if an override is not provided.

Before defaulting to the `created_at` timestamp we should use `finalized_at` if available

![Screen Shot 2021-10-15 at 10 40 29 PM](https://user-images.githubusercontent.com/4914611/137551038-2b9415be-6df3-4426-b6e3-ae81b0ee343f.png)

PR #558, opened 2021-11-09
Use finalize_at instead of updating the created at timestamp on invoice webhooks.

mecoccaro:
**Issue #546**
**Description**

When a Stripe invoice is finalized and set to ready_to_allocate to be triggered by the webhook to create a Payment, the logic correctly brings in the date_incurred data when creating or updating the Trinary Payment. If provided in the metadata, it uses the override value or defaults to the Invoice creation timestamp if an override is not provided.

Before defaulting to the created_at timestamp we should use finalized_at if available

**Implementation proof**
https://www.loom.com/share/1cc940561a344743824c2d40230fb269

### DIFF
--- a/api/modules/budgeting.js
+++ b/api/modules/budgeting.js
@@ -32,7 +32,7 @@ const budgeting = module.exports = (() => {
             const dateIncurredOverride = stripeInvoice.metadata?.date_incurred
                 ? moment(stripeInvoice.metadata.date_incurred, 'YYYY-MM-DD')
                 : stripeInvoice.status_transitions?.finalized_at
-                    ? moment.unix(stripeInvoice.status_transitions.finalized_at, 'x').utc()
+                    ? moment.unix(stripeInvoice.status_transitions.finalized_at, 'X')
                     : moment(stripeInvoice.created, 'X')
 
             const datePaidOverride = stripeInvoice.metadata?.date_paid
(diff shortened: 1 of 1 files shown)


sofiaromorales (approved):

api/modules/budgeting.js
```diff
@@ -142,6 +145,8 @@ const budgeting = module.exports = (() => {
 
             if (dateIncurredOverride) {
                 updatedAttributes.date_incurred = dateIncurredOverride
+            } else if (finalizedAt) {
+                updatedAttributes.date_incurrred = moment(finalizedAt).format()
```
otech47, line 149: Fix typo: `date_incurrred` to `date_incurred`

otech47: Let's also format it to be `format('YYYY-MM-DD')` since that is the same format used when filling out `date_incurred` from the Stripe metadata

api/modules/budgeting.js
```diff
@@ -31,7 +31,9 @@ const budgeting = module.exports = (() => {
         if (client) {
             const dateIncurredOverride = stripeInvoice.metadata?.date_incurred
                 ? moment(stripeInvoice.metadata.date_incurred, 'YYYY-MM-DD')
-                : moment(stripeInvoice.created, 'X')
+                : stripeInvoice.status_transitions?.finalized_at
+                    ? moment.unix(stripeInvoice.status_transitions.finalized_at, 'x').utc()
```
otech47, line 35: Change to simply `moment(stripeInvoice.status_transitions.finalized_at, 'X')` since it should just use the same formatting as `moment(stripeInvoice.created, 'X')` which was working fine before. Even if it seems to work there's no need to introduce different functions for this since Stripe will give the exact same format for both `finalized_at` and `created`

otech47 (changes requested):

otech47 (approved):

Issue #555, opened 2021-10-25
Allocations cannot be edited if the dates have changed


otech47:
The Save button is disabled even if you change the start/end dates. Should be able to edit and save the allocation with different dates

PR #560, opened 2021-11-12
Allocations cannot be edited if the dates have changed

mecoccaro:
**Issue #555**
**Description**

The Save button is disabled even if you change the start/end dates.

**Implementation proof**
https://www.loom.com/share/267658598e934d3cba67b97d973ee0fe?sharedAppSource=personal_library

### DIFF
--- a/src/components/AllocationOverview.js
+++ b/src/components/AllocationOverview.js
@@ -232,7 +232,8 @@ const AllocationOverview = (props) => {
         updatedAllocationRate.total_amount == allocationInfo.amount &&
         updatedAllocationRate.hourly_rate == allocationInfo.rate.hourly_rate &&
         updatedAllocationPayment?.id == allocationInfo.payment?.id &&
-        updatedAllocationEndDate == allocationInfo.end_date
+        updatedAllocationEndDate == allocationInfo.end_date ||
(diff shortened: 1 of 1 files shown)


src/components/AllocationOverview.js
```diff
@@ -231,9 +231,9 @@ const AllocationOverview = (props) => {
     const editButtonDisabled = (
         updatedAllocationRate.total_amount == allocationInfo.amount &&
         updatedAllocationRate.hourly_rate == allocationInfo.rate.hourly_rate &&
-        updatedAllocationPayment?.id == allocationInfo.payment?.id
+        updatedAllocationPayment?.id == allocationInfo.payment?.id &&
+        updatedAllocationEndDate == allocationInfo.end_date
```
otech47, line 235: Don't we also need to add `start_date` to this condition? @mecoccaro 

the test case didn't show when only the start date changes...

If everything except the start date is the same it seems this code would still disable the button right? Which is not the desired behavior. Am i missing something?

mecoccaro: As it is right now, when you modify either start or end date the button will be enable because the calendar functionality changes the entire date, no matter if you just selected the start or end date.

otech47 (approved):

Issue #361, opened 2021-03-10
Edit Project is not handling the Toggl URL properly


otech47:
https://www.loom.com/share/a5845393e43b4ba09c85900a78ba3acc

Tested from both Edit Project and Add Project forms neither seems to be working. It's possible maybe the Toggl validation rules need to be clarified in the error message?

otech47:
I think maybe this wasn't resolved properly as it is still happening in production after the latest fixes:

https://www.loom.com/share/817445bd78054322a6aac0f47c19006d

It is still confusing to the user and they would not know about the validation rules

PR #571, opened 2021-11-25
Toggl url validation rules explained

mecoccaro:
**issue #361**
**Description**

When the user set the toggl url incorrectly it is still confusing and they would not know about the validation rules.

Updated the error popup message.

imagen


sofiaromorales (changes requested):

src/components/AddProjectForm.js
```diff
@@ -118,7 +118,8 @@ const AddProjectForm = (props) => {
         }
         if (projectToggl) {
             if (!verifyTogglURL(projectToggl)) {
-                setCreateProjectError('The Toggl URL is invalid')
+                setCreateProjectError('The Toggl URL is invalid.')
+                setCreateProjectError('Please follow the following format https://track.toggl.com/###/projects/###')
```
otech47, line 122: Let's match this with https://github.com/setlife-network/trinary/pull/571/files#diff-ebbc47c19924a093eb3b1dc5e7b0f9948b71855eb4f834a4ec691758a6cd3508R102 by creating a constant `INVALID_TOGGL_URL_ERROR_MESSAGE` and importing it in both files

otech47 (approved):

Issue #595, opened 2021-12-31
Add extra bottom margin to the list of payment allocations on the Project Budgeting page so that the bottom most tiles don't get cut off


otech47:
Add extra bottom margin to the list of payment allocations on the Project Budgeting page so that the bottom most tiles don't get cut off
![Screen Shot 2021-12-22 at 11 10 44 AM](https://user-images.githubusercontent.com/4914611/147122074-4d50535a-4de0-4e77-beeb-d17af8206404.png)

_Originally posted by @otech47 in https://github.com/setlife-network/trinary/discussions/588#discussioncomment-1859272_

PR #598, opened 2022-01-04
Add extra bottom margin to the list of payments allocations on the Project Budgeting page

Ovifer13:
**Issue #595**
**Description**
Add extra bottom margin to Allocations in `ProjectPage` so the tile doesn't get cut off with the navigation buttons.

**Proof**
image


src/pages/ProjectDetailPage.js
```diff
@@ -3,7 +3,8 @@ import { Route, Redirect } from 'react-router-dom'
 import {
     BottomNavigation,
     BottomNavigationAction,
-    Grid
+    Grid,
+    Box
```
otech47, line 7: Box is not needed here? @vdegandesa

src/pages/ProjectDetailPage.js
```diff
@@ -32,7 +33,12 @@ class ProjectDetailPage extends React.Component {
 
         return (
             <Grid container justify='center' className='ProjectDetailPage'>
-                <Grid item xs={10} md={6}>
+                <Grid 
+                    item 
+                    xs={10} 
+                    md={6} 
+                    className='GridProjectDetailPage'
```
otech47, line 40: When adding class names to DOM elements that aren't a parent container for a React component, we should use a different naming convention so it doesn't cause confusion reading SCSS files that might refer to a capitalize camelcase name 

Suggestion: rename to kebab-case `grid-project-detail-page` or `grid-container`

otech47 (changes requested):

src/styles/ProjectDetailPage.scss
```diff
@@ -5,6 +5,9 @@
         width: 100%;
         z-index: 2
     }
+    .GridProjectDetailPage {
+        margin-bottom: 2rem;
+    }
```
otech47, line 10: Suggestion:

```
> .MuiGrid-item {
    margin-bottom: 2rem;
}
```

otech47 (changes requested):

src/styles/ProjectDetailPage.scss
Ovifer13: If we use `.MuiGrid-item` this happens 
image

otech47 (approved):

Issue #591, opened 2021-12-31
Styling updates to Client Detail Page


otech47:
![Screen Shot 2021-12-20 at 11 38 53 AM](https://user-images.githubusercontent.com/4914611/146802813-f00cf0b2-98bd-4895-b477-8b69a1f194fa.png)

_Originally posted by @otech47 in https://github.com/setlife-network/trinary/discussions/588#discussioncomment-1848285_

sofiaromorales:
@Ovifer13 are you working on this?

PR #600, opened 2022-01-04
Styling updates to Client Detail  Page

Ovifer13:
**Issue #591**
**Description**
Styling changes for the `ClientDetail` Page:
- [x] Reduce project name font size to match Client name in header.
- [ ] Reduce horizontal and vertical padding of both containers.
- [ ] Keep alignment but reduce padding.


Ovifer13:
**Update checkbox 1**
The client name was also reduced to match with the project name size because it was bigger than the new size.
image

sofiaromorales:
@Ovifer13 are you working on this?

PR #606, opened 2022-01-15
Issues pagination

sofiaromorales:


sofiaromorales:
Screen Shot 2022-01-21 at 12 04 07 AM

api/schema/resolvers/IssueResolver.js
```diff
@@ -13,8 +16,22 @@ module.exports = {
...
+            }
+            if (last_30_days_only) {
+                issuesConditions.created_at = {
+                    [Op.gte]: `${moment().subtract(30, 'days').toDate()}`
+                }
+            }
+            return models.Issue.findAll({
+                where: issuesConditions,
+                limit: limit,
+                offset: offset,
```
otech47, line 31: Shouldn't we have defaults here if these are not provided? They would be `undefined` no?

otech47: @Ovifer13 can you take a look at this PR and finish it if you have time?

Ovifer13:
@otech47 I tested this PR and it's working. Didn't understand the last comment you did, if you could explain me 
> Shouldn't we have defaults here if these are not provided? They would be `undefined` no?

Issue #610, opened 2022-01-18
Fix empty client email field


otech47:
A couple problems here

1. Should be able to create a client with an empty email field (but the value in the DB should be NULL not an empty string)

![Screen Shot 2022-01-18 at 10 20 53 AM](https://user-images.githubusercontent.com/4914611/149965770-3081047a-e2c2-4218-b07c-c9776afb3624.png)

2. When updating a client who does have an email and clearing out the email field, when saving to the DB it should be a NULL value instead of an empty string

![Screen Shot 2022-01-18 at 10 22 49 AM](https://user-images.githubusercontent.com/4914611/149965984-85deee6d-196a-4b4c-8d8f-5e9f7e1f6a9e.png)

The reason for this is because now with the unique email client constraint from #173 I believe is causing the 1st bug because an empty string will count as a unique email and block updates to the DB with empty emails

PR #611, opened 2022-01-18
Fix empty client Email

mecoccaro:
**Issue #610**
**Description**

- [x] Should be able to create a client with an empty email field (but the value in the DB should be NULL not an empty string)

- [x] When updating a client who does have an email and clearing out the email field, when saving to the DB it should be a NULL value instead of an empty string

**Implementation proof**
https://www.loom.com/share/12b8be21109447fbb99da8b6826aab97
imagen

### DIFF
--- a/api/handlers/stripe.js
+++ b/api/handlers/stripe.js
@@ -15,7 +15,7 @@ const stripeHandler = module.exports = (() => {
         const { email, name } = params
 
         return stripeClient.customers.create({
-            email,
+            email: email ? email : null,
             name,
         })
     }
--- a/api/modules/clientManagement.js
+++ b/api/modules/clientManagement.js
@@ -37,6 +37,10 @@ const clientManagement = module.exports = (() => {
         try {
             const { createFields } = params
 
+            if (!createFields.email) {
+                createFields.email = null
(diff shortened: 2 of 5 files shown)


sofiaromorales (commented):
@mecoccaro I still cant edit the email into an empty field

mecoccaro:
@sofiaromorales I tested again trying to delete the email from an existing client and it worked, if you can, try again and let me know. Also if @otech47 or @Ovifer13 can test would it be great!

Ovifer13:
@mecoccaro I did a test with a user that already had an email and I was able to edit it with the mail field empty.

otech47:
@Ovifer13 can you take another look at this and make sure it works as expected?

Ovifer13:
@otech47 I'm not getting null on the email field after I edit it as an empty field, want me to fix this?

Ovifer13:
**Implementation proof**
<a href="https://www.loom.com/share/a81bcf81a9b44bacbfad3f47ef36161c">
    <p>Fix empty client Email after editing - Watch Video</p>
    
  </a>

otech47 (approved):

Issue #562, opened 2021-11-15
Display github user image instead of github username in author and assignee inside IssueTile


Ovifer13:
Instead of displaying Github usernames, it would be better to display the GitHub user image inside a circle, like in AddProjectPage.

![image](https://user-images.githubusercontent.com/86666889/141845626-c7b6a503-e8f5-4d1b-a2a3-d57b7d7e027e.png)

Ovifer13:
@otech47 @sofiaromorales Can we add the avatar URL to the contributor table? That's one option I was thinking about of how to do this Issue. Do you guys think of a better solution?

PR #615, opened 2022-01-27
Display user avatar in IssueTile

Ovifer13:
**Issue #562**
**Description**
Instead of displaying the `github_handler` in `IssueTile`, the _avatar_ of the user should be displayed.

- [x] Create avatar_url database column.
- [x] Create a migration to add the column.
- [x] Use the avatar column to get the avatar_url in the `IssueTile` component.


Ovifer13:
**Update**
image

otech47 (approved):

PR #627, opened 2022-04-15
generateBitcoinInvoiceFromPayment mutation added

rishkwal:


api/handlers/btcPayServer.js
```diff
@@ -0,0 +1,54 @@
+const axios = require('axios')
+const {
+    BTCPAYSERVER
+} = require('../config/credentials')
+
+const BTCPayServerAPIKey = BTCPAYSERVER.API_KEY;
+const BTCPayServerStoreId = BTCPAYSERVER.STORE_ID;
```
otech47, line 7: Rename to `BTCPAYSERVER_API_KEY` and `BTCPAYSERVER_STORE_ID` for constants like this

api/handlers/btcPayServer.js
```diff
@@ -0,0 +1,54 @@
...
+                'BTC', 'BTC-LightningNetwork'
+            ],
+            'defaultPaymentMethod': 'BTC-LightningNetwork',
+            'expirationMinutes': 5,
+            'monitoringMinutes': 5,
+            'paymentTolerance': 100,
+        },
+        'amount': amount,
+        'currency': 'sats',
+        'additionalSearchTerms': [
```
otech47, line 39: Remove all single quotes around property fields

api/handlers/btcPayServer.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+    return axios.post(`https://btcpayserver.setlife.tech/api/v1/stores/${BTCPayServerStoreId}/invoices`, body, config).then(res => res.data);
+}
+
+const getAllInvoices = () => {
+    axios.get(`https://btcpayserver.setlife.tech/api/v1/stores/${BTCPayServerStoreId}/invoices`, config).then(res => res.data);
+}
+
+const getInvoiceById = (invoiceId) => {
+    return axios.get(`https://btcpayserver.setlife.tech/api/v1/stores/${BTCPayServerStoreId}/invoices/${invoiceId}`, config).then(res => res.data);
```
otech47, line 51: Let's fix a few things here as coding standard:

- Prefer `async/await` instead of `.then`
- use newlines for better code readability
- Create a `BTCPAYSERVER_API_ROOT` constant for better resuability

```
const BTCPAYSERVER_API_ROOT = `https://btcpayserver.setlife.tech/api/v1`

const getAllInvoices = async () => {
    const response = await axios.post(
        `${BTCPAYSERVER_API_ROOT}/stores/${BTCPayServerStoreId}/invoices`,
        body,
        config
    )
    
    return response.data
}
```

api/handlers/btcPayServer.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+module.exports = { checkInvoiceExpired, createBitcoinInvoice, getAllInvoices, getInvoiceById };
```
otech47, line 54: Change to 1 line per function for better readability

api/schema/resolvers/PaymentResolver.js
```diff
@@ -77,6 +78,32 @@ module.exports = {
         deletePaymentById: (root, { id }, { models }) => {
             return models.Payment.destroy({ where: { id } })
         },
+        generateBitcoinInvoiceFromPayment: async (root, { paymentId }, { models }) => {
+            
+            const Payment = await models.Payment.findByPk(paymentId)
```
otech47, line 83: Avoid capitalized 1st letters for basic variable names

api/schema/resolvers/PaymentResolver.js
```diff
@@ -77,6 +78,32 @@ module.exports = {
...
+            const { amount, client_id, date_paid, external_uuid, external_uuid_type } = Payment.dataValues
+
+            const client = await models.Client.findByPk(client_id)
+            
+            const isClientCurrencyBtc = client.dataValues.currency === 'BTC' || client.dataValues.currency === 'SATS'
+            if (!isClientCurrencyBtc) throw 'Client\'s currency is not Bitcoin'
```
otech47, line 89: Easier way to handle escape characters: 

```
if (!isClientCurrencyBtc) throw `Client's currency is not Bitcoin`
```

api/schema/resolvers/PaymentResolver.js
```diff
@@ -77,6 +78,32 @@ module.exports = {
...
+
+            const isInvoiceExpired = external_uuid && external_uuid_type === 'bitcoin' && await checkInvoiceExpired(external_uuid)
+            if (external_uuid && date_paid && !isInvoiceExpired) throw 'An active invoice already exists'
+
+            const amountInSats = client.dataValues.currency === 'BTC' ? amount * 1000000 : amount / 100;
```
otech47, line 94: Probably want to guard this from floating point errors

```
const amountInSats = client.dataValues.currency === 'BTC'
    ? Number((amount * 1000000).toFixed(0))
    : Number((amount / 100).toFixed(0))
```

Also why are we dividing by 100 and multiplying by 1000000? 100M sats = 1 BTC

api/schema/resolvers/PaymentResolver.js
```diff
@@ -77,6 +78,32 @@ module.exports = {
...
+            const newInvoice = await createBitcoinInvoice(amountInSats)
+            return models.Payment.update({
+                external_uuid: newInvoice.id,
+                external_uuid_type: 'bitcoin'
+            }, 
+            { where: 
+                {
+                    id: paymentId
+                } 
+            })
```
otech47, line 105: This notation is a bit wonky, should be
```
return models.Payment.update(
    {
        external_uuid: newInvoice.id,
        external_uuid_type: 'bitcoin'
    },
    {
        where: {
            id: paymentId
        }
    }
)
```

or perhaps
```

return models.Payment.update(
    {
        external_uuid: newInvoice.id,
        external_uuid_type: 'bitcoin'
    },
    {
        where: { id: paymentId }
    }
)
```

this makes it more clear where the parameters of the function start and end

otech47, line 105: Let's also refactor all of this logic into an API module so we don't tons of logic in our GraphQL resolvers

As a separation of concerns I would recommend resolvers be reserved for simple logic that is needed for tweaking incoming arguments before sending it to a function in `/api/modules/...` or perhaps tweaking the outgoing response but most of the core logic, especially making requests to other files like the `btcPayServer` handler

This way our resolvers depend on API modules and nothing else and the modules can have further dependencies

otech47 (changes requested):

api/schema/resolvers/PaymentResolver.js
rishkwal: The data type for `amount` in the database is Integer and the input for payment amount allows up to two decimal places. So the database actually stores 100 times the actual amount to store it as an Integer. These calculations are for adjusting back the amount by dividing it by 100.

otech47: Ah okay that's right, could be worth adding a comment or describing with variables:

```
const convertedAmount = Number((amount / 100).toFixed(0))

const amountInSats = client.dataValues.currency === 'BTC'
    ? Number((convertedAmount * 100000000).toFixed(0)) // 100M sats = 1 BTC
    : convertedAmount
```

rishkwal:
Features added: https://www.loom.com/share/c0f79e625c394bde8853f8502fbf29af

otech47 (approved):

Issue #637, opened 2022-05-11
When an user empties the date field and tries to save changes without a date the app should give feedback of the error


Ovifer13:
<a href="https://www.loom.com/share/057aafe176754c6ebb5d91816cf63719">
    <p>Project Trinary - Error when saving payment and editing project - Watch Video</p>
    <img style="max-width:300px;" src="https://cdn.loom.com/sessions/thumbnails/057aafe176754c6ebb5d91816cf63719-with-play.gif">
  </a>

Ovifer13:
@otech Do trinary uses toast notifications? I'm searching for the _toast_ keyword and can't find anything.

PR #642, opened 2022-05-18
When an user empties the date field and tries to save changes without a date the app should give feedback of the error

Ovifer13:
**Issue #637**
- [x] When adding a payment if the amount field is empty disable the **Add Payment** button
- [x] When adding a payment if the incurred date is empty, disable the **Add Payment** button
- [x] When editing a payment if the amount field is empty disable the **Save Payment** button
- [x] When editing a payment if the incurred date is empty disable the **Save Payment** button
- [x] When editing a project if the amount field is empty disable the **Save Project** button
- [x] When editing a project if the incurred date is empty disable the **Save Project** button
- [x] Handle errors so the app doesn't crash and show the error to the user in all components

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "react-dom": "^16.14.0",
     "react-router-dom": "^5.2.0",
     "react-scripts": "5.0.0",
-    "react-toast-notifications": "^2.5.1",
     "sequelize": "^6.3.5",
     "sequelize-cli": "^6.2.0",
     "stripe": "^8.150.0",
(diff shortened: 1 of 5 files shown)


Ovifer13:
@otech47  I was able to disable the button if there is no amount or payment. Should I also add a toast to notify the user in case there's an error?

otech47 (changes requested):
@Ovifer13 can you revert the react-toast-notifications commits?

I didn't realize but we are already using a toasts implementation with the [Snackbar](https://mui.com/material-ui/react-snackbar/) component [here](https://github.com/setlife-network/trinary/pull/642/files#diff-7394fc6765254ae6d43abe2490d6149c8b353ee258e1fb15e37432635253a2f1R199) so it doesn't really make sense to have 2 different implementations for displaying errors

otech47 (approved):

Issue #646, opened 2022-05-20
Allow users to edit and delete payments from the Project Detail page


otech47:
- Do not allow deleting if the Payment has Allocations

Show "This payment cannot be deleted because it has linked Allocations" as a tooltip
![Screen Shot 2022-05-20 at 4 22 59 PM](https://user-images.githubusercontent.com/4914611/169548600-9e207b0e-9de7-46f9-95a9-b69c73cd74e0.png)

Ovifer13:
**Tested in Stagging**
- The delete button is disabled when there are allocations.
- If the payment doesn't has allocations the delete button works and the user can proceed to delete a payment.

PR #653, opened 2022-06-07
Do not allow deleting if the Payment has Allocations

Ovifer13:
**Issue #646**
- [x] Show Tooltip when hovering the mouse on the delete button _"This payment cannot be deleted because it has linked Allocations"_
- [x] Disable delete button when the payment has allocations
- [x] Enable delete button when there are no allocations


Ovifer13:
**Implementation proof**
image

otech47 (approved):

Issue #640, opened 2022-05-17
Update Rates data model


otech47:
This is a backend only issue and any UI changes will be in a separate issue

## Roles
- [x] Create a new `roles` table with fields:

1. id
2. name - VARCHAR(255)
3. created_at - default
3. updated_at - default

This table will have future uses but for now will just have 1-to-many with the Rates table

## Rates
Anywhere in the API that `hourly_rate` and `total_expected_hours` can be provided for a Rate+Allocation (`createRate`, etc) we should allow these new fields to be provided

- [x] Add all of these new fields which should be NULLABLE and should default to NULL
1. minimum_expected_hours - INT(11)
2. maximum_expected_hours - INT(11)
3. minimum_hourly_rate - VARCHAR(15)
4. maximum_hourly_rate - VARCHAR(15)
5. role_id - INT(11) - default to NULL

- [x] Change `total_expected_hours` datatype to match new minimum/maximum fields
- [x] Change `hourly_rate` datatype to match new minimum/maximum fields

PR #656, opened 2022-06-14
Update Rates data model

Ovifer13:
**Issue #640**
- [x] Create new `roles` table with fields:
1. id
2. name -VARCHAR(255)
3. create_at - default
4. updated_at default

- [x] Add new fields to `rates` table, all fields should be **NULLABLE** and default to _NULL_
1. minimum_expected_hours - INT(11)
2. maximum_expected_hours(11)
3. minimum_hourly_rate - VARCHAR(15)
4. maximum_hourly_rate VARCHAR(15)
5. role_id - INT(11) - default to NULL

- [x] Change `total_expeted_hours` datatype to match new minimum/maximum fields in rates table
- [x] Change `hourly_rate` datatype to match new minimum/maximum fields
- [x] Anywhere in the API that `hourly_rate` and `total_expected_hours` can be provided for a Rate+Allocation, allow these new fields to be provided

### DIFF
--- a/api/migrations/20220609175823-createRoles.js
+++ b/api/migrations/20220609175823-createRoles.js
@@ -1,6 +1,6 @@
 module.exports = {
     up: async (queryInterface, Sequelize) => {
-        return queryInterface.createTable('Role', {
+        return queryInterface.createTable('Roles', {
             id: {
                 type: Sequelize.DataTypes.INTEGER,
                 primaryKey: true,
@@ -15,6 +15,6 @@ module.exports = {
     },
 
     down: async (queryInterface, Sequelize) => {
-        return queryInterface.dropTable('Contributions');
+        return queryInterface.dropTable('Roles');
     }
 };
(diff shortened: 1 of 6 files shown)


Ovifer13:
Also the API gives feedback of the new fields
image

api/migrations/20220609175823-createRoles.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        return queryInterface.createTable('Role', {
```
otech47, line 3: The table should be called `Roles` to match the pluralized naming convention of all other tables

api/migrations/20220609175823-createRoles.js
```diff
@@ -0,0 +1,20 @@
...
+            },
+            name: {
+                type: Sequelize.DataTypes.STRING,
+                allowNull: false
+            }
+        });
+    },
+
+    down: async (queryInterface, Sequelize) => {
+        return queryInterface.dropTable('Contributions');
```
otech47, line 18: Fix bad down migration that is removing the wrong table

otech47 (changes requested):

otech47 (approved):

PR #660, opened 2022-06-24
Disabling change of Payment details after BTC invoice has been paid

rishkwal:


rishkwal:
### 1. Updating the `date_paid`
Here is a payment with `id=147`, `date_paid=null` and an unpaid BTCPS invoice

Screen Shot 2022-07-01 at 3 46 53 PM

The invoice is then paid

Screen Shot 2022-07-01 at 3 51 20 PM

A payload is received from the webhook that automatically updates the `date_paid` to the date on which the invoice is paid

Screen Shot 2022-07-01 at 3 54 55 PM

### 2. Whenever a user tries to edit a Payment with an existing invoice, if the invoice is already settled, an error is thrown.
`No further changes can be made to this payment`
Screen Shot 2022-07-01 at 3 57 20 PM

### 3. Prevent user from creating an invoice when `date_paid!=null`
An error with the message `Payment has already been made` is thrown
Screen Shot 2022-07-01 at 3 58 11 PM

otech47 (approved):

PR #671, opened 2022-07-09
BTC invoice generation bug resolved

rishkwal:
**Change 1**: `amountUpdated` checks if the amount for the current BTCPS invoice for the Payment is different from the Payment amount. To retrieve the BTCPS Invoice amount a request is sent to BTC Pay Server and `external_uuid` has the Invoice Id. Before sending the request, a check for whether the `external_uuid` exists was absent. Therefore, new Payments were sending requests with `external_uuid=null` and getting an error whereas Payments with pre-existing invoice could generate a new Invoice. A check for `external_uuid` was added for `amountUpdated`.

**Change 2**: `date_paid` need not exist to check if `An active invoice already exists`

### DIFF
--- a/api/modules/paymentManagement.js
+++ b/api/modules/paymentManagement.js
@@ -21,11 +21,6 @@ const paymentManagement = module.exports = (() => {
         return invoice.checkoutLink
     }
 
-    const getBitcoinInvoiceAmount = async (invoiceId) => {
-        const invoice = await getInvoiceById(invoiceId)
-        return Number(invoice.amount)
-    }
-
     const processBitcoinInvoiceCreation = async (paymentId) => {
         const payment = await db.models.Payment.findByPk(paymentId)
         const { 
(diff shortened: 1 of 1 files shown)


rishkwal:
**Implementation Proof**
trinary_payment

api/modules/paymentManagement.js
```diff
@@ -54,9 +54,10 @@ const paymentManagement = module.exports = (() => {
             ? Number((convertedAmount * 100000000).toFixed(0)) // 100M sats = 1 BTC
             : convertedAmount
         
-        const amountUpdated = amountInSats != await getBitcoinInvoiceAmount(external_uuid)
+        const amountUpdated = external_uuid && 
+        ( amountInSats != await getBitcoinInvoiceAmount(external_uuid) )
         
-        if (external_uuid && date_paid && !isInvoiceExpired && !amountUpdated) throw new Error(`An active invoice already exists`)
+        if (external_uuid && !isInvoiceExpired && !amountUpdated) throw new Error(`An active invoice already exists`)
```
otech47, line 60: Let's wrap all logic that expects `external_uuid` to have a value so that the 1st time we create an invoice is a bit cleaner code-wise

----------

NIT on indenting pattern:

```
const amountUpdated = (
    external_uuid && 
    amountInSats != await getBitcoinInvoiceAmount(external_uuid)
)
```

otech47 (approved):

Issue #672, opened 2022-07-11
App breaks when adding an Allocation


Ovifer13:
When adding an allocation the app breaks and the following error is shown in the console
![image](https://user-images.githubusercontent.com/86666889/178312368-26fc8c7c-dde6-4e2d-9324-e9ad44a70518.png)

PR #673, opened 2022-07-12
Fix app break and add missing variables

Ovifer13:
**Issue #672**
- [x] Fix app breaks when trying to add an allocation
- [x] Add missing variables to create new rate fields in the database


Ovifer13:
**Implementation proof**
image

otech47 (approved):

Issue #645, opened 2022-05-20
Link multi-project Allocations to Client Payments


otech47:
Blocked by https://github.com/setlife-network/trinary/issues/644

- When the user clicks the `$500 allocated to 2 other project(s)` text on a PaymentTile on the ProjectDetail page, then the Client Detail page should appear, automatically scroll the page down to the correct PaymentTile, AND automatically expand the Payment

PR #674, opened 2022-07-13
Link multi-project Allocations to Client Payments

Ovifer13:
**Issue #645**
- [x] Make the text `$500 allocated to 2 other project(s)` clickable
- [x] After the text was clicked, redirect to the client page and expand the payment from the project that was clicked

### DIFF
--- a/src/components/PaymentTile.js
+++ b/src/components/PaymentTile.js
@@ -38,9 +38,11 @@ const PaymentTile = (props) => {
     const {
         client,
         payment,
-        project
+        project,
+        active
     } = props
 
+    localStorage.removeItem('openPayment')
     const formattedDatePaid = moment.utc(parseInt(payment.date_paid, 10)).format('MM/DD/YYYY')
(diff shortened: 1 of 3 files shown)


src/components/PaymentTile.js
```diff
@@ -85,6 +88,13 @@ const PaymentTile = (props) => {
     const handleEditPayment = () => {
         history.push(`/clients/${client.id}/payments/${payment.id}/update`)
     }
+    const handleRedirect = () => {
+        history.push(`/clients/${client.id}`)
+        localStorage.setItem('openPayment', payment.id)
```
otech47, line 93: Instead of `localStorage` we should just use the URL params so you would have

```
history.push(`/clients/${client.id}?paymentId=${payment.id}`)
```

And then you can just pull it out of the URL in [PaymentsList.js](https://github.com/setlife-network/trinary/pull/674/files#diff-c1738c305cf76b78901c816f6b869a371c366ddd6ecdcfa20f669806574bc7b3R25)

otech47 (changes requested):

otech47 (approved):

PR #681, opened 2022-09-13
Add a link to the Client Detail page from the Project Detail page

### LINKED ISSUE #680: Add a link to the Client Detail page from the Project Detail page
Screen Shot 2022-08-16 at 7 28 41 PM

Ovifer13:
**Issue #680**

- [x] Add redirect link to client text
- [x] Add a pointer style
- [x] Add color change on hover

**Proof**
https://www.loom.com/share/d4ded633a8dd4b658394fe5d755dee6d

### DIFF
--- a/src/components/ProjectSummary.js
+++ b/src/components/ProjectSummary.js
@@ -67,7 +67,7 @@ const ProjectSummary = (props) => {
                                 <Grid xs={10} align='left'>
                                     <Typography 
                                         variant='' 
-                                        className='redirect'
+                                        className='grey-link'
                                         onClick={redirectToClient}
                                     >
                                         <span>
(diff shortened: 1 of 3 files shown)


src/components/ProjectSummary.js
```diff
@@ -40,6 +43,10 @@ const ProjectSummary = (props) => {
         currencyInformation: currencyInformation
     })
 
+    const redirectToClient = () => {
```
otech47: the word "redirect" is confusing because it is a term for redirecting old web URLs to new ones

change this to `goToClientPage`

src/styles/ProjectSummary.scss
```diff
@@ -7,6 +7,11 @@
 
 @media screen and (min-width: $sm) {
     .ProjectSummary {
-
+        .redirect {
```
otech47: These styles are pretty generic. If there is really nothing built into Material UI that adds a cursor pointer and changes to grey on hover we should at least make a `grey-link` class in a global style that can be re-used

otech47 (changes requested):

otech47 (approved):

Issue #705, opened 2022-11-02
Define and implement API changes to facilitate allocation countering/rate negotation


otech47:

stark-3k:
**Possible Negotiation Database Implementation**
In the `Allocations` table, we have a `status` column with four options: `accepted`,`rejected`,`pending` or `countered`
Assuming we don't need to maintain the history of the allocation:
-  If the status of the Allocation is `pending`, the allocation has been proposed by the Project Admin and can be `accepted`,`rejected` or `countered` by the Contributor
- If the status of the Allocation is `countered`, the allocation has been proposed by the Contributor and can be `accepted`,`rejected` or `countered` by the Project Admin

otech47:
We should also change all places where allocations are fetched (ex: `Contributor.allocations`, `Project.allocations` or `getAllocations`, etc) to filter by `status`

```
getContributor {
  allocations(
    filter: {
      status: ['accepted', 'rejected']
    }
  ) {
    amount
    status
  }
}

otech47:
@rishkwal we also want a `proposed_by_contributor_id` so that we know who originally proposed the Allocation

sofiaromorales:
> proposed_by_contributor_id



> @rishkwal we also want a `proposed_by_contributor_id` so that we know who originally proposed the Allocation

This should be nullable

stark-3k:
> > proposed_by_contributor_id
> 
> > @rishkwal we also want a `proposed_by_contributor_id` so that we know who originally proposed the Allocation
> 
> This should be nullable

Thanks for pointing out. #725 will rectify the mistake.

PR #725, opened 2022-11-18
`proposed_by_contributor_id` can be set to null

stark-3k:
https://github.com/setlife-network/trinary/issues/705#issuecomment-1320179824


stark-3k:
Merging #727 will resolve

PR #739, opened 2022-11-23
Allow creation of payments without client Id

stark-3k:
- Eliminated the need of `client_id` to create Payments
- Added `currency` field to `Payment`
- Added `contributor_id` to `Payment`
- Enabled generation of BTC invoices without needing a Client

### DIFF
--- a/api/migrations/20221123064500-addCurrencyAndContributorToPayment.js
+++ b/api/migrations/20221123064500-addCurrencyAndContributorToPayment.js
@@ -5,7 +5,7 @@ module.exports = {
           queryInterface.addColumn('Payments', 'currency', {
             type: Sequelize.DataTypes.STRING
           }, { transaction: t }),
-          queryInterface.addColumn('Paymetns', 'contributor_id', {
+          queryInterface.addColumn('Payments', 'contributor_id', {
             type: Sequelize.DataTypes.INTEGER,
             allowNull: true,
             references: {
--- a/api/migrations/20221123064500-addCurrencyAndContributorToPayment.js
+++ b/api/migrations/20221123064500-addCurrencyAndContributorToPayment.js
@@ -1,0 +1,31 @@
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    return queryInterface.sequelize.transaction(t => {
+      return Promise.all([
+          queryInterface.addColumn('Payments', 'currency', {
+            type: Sequelize.DataTypes.STRING
+          }, { transaction: t }),
+          queryInterface.addColumn('Paymetns', 'contributor_id', {
+            type: Sequelize.DataTypes.INTEGER,
+            allowNull: true,
+            references: {
+              model: 'Contributors',
+              id: 'key'
+            }
+          }, { transaction: t })
+        ])
+    })
+  },
+
+  async down (queryInterface, Sequelize) {
+    return queryInterface.sequelize.transaction(t => {
+      return Promise.all([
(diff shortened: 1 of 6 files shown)


api/modules/paymentManagement.js
```diff
@@ -22,29 +22,41 @@ const paymentManagement = module.exports = (() => {
...
+            paymentCurrency = client.dataValues.currency
+            isClientCurrencyBtc = 
+            paymentCurrency === `BTC` ||
+            paymentCurrency === `SATS`
+        }
+
+        if (currency) {
+            isClientCurrencyBtc = currency === `BTC` || currency === `SATS`
+            paymentCurrency = currency
+        }
```
sofiaromorales: @rishkwal if the currency is not nullable this will always be true right? Instead of `if` maybe `else` makes more sense so we don't overwrite the other logic

api/schema/types/PaymentType.js
```diff
@@ -7,7 +7,9 @@ module.exports = gql`
         amount: Int!
         date_incurred: String!
         date_paid: String
-        client_id: Int!
+        client_id: Int
+        contributor_id: Int
+        currency: String
```
sofiaromorales, line 12: You made currency not nullable in the model so we should enforce it at the API level

sofiaromorales (commented):
I'm a but confused about some of this logic

api/migrations/20221123064500-addCurrencyAndContributorToPayment.js
```diff
@@ -0,0 +1,31 @@
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    return queryInterface.sequelize.transaction(t => {
+      return Promise.all([
+          queryInterface.addColumn('Payments', 'currency', {
+            type: Sequelize.DataTypes.STRING
+          }, { transaction: t }),
+          queryInterface.addColumn('Paymetns', 'contributor_id', {
```
sofiaromorales: fix typo

sofiaromorales (changes requested):

api/modules/paymentManagement.js
stark-3k: You're correct, this is always going to be true.
We can either make `currency` nullable
OR
Maybe we can remove the currency of `client` from the logic and derive the currency only from the Payment
```suggestion
            isClientCurrencyBtc = currency === `BTC` || currency === `SATS`
            paymentCurrency = currency
```
What do you suggest?

sofiaromorales: I would go with this approach to make it more flexible. Even we can default to the project currency if none of those cases are met
```
 if (currency) {
            isClientCurrencyBtc = currency === `BTC` || currency === `SATS`
            paymentCurrency = currency
        }
else if (client_id) {
            const client = await db.models.Client.findByPk(client_id)
            paymentCurrency = client.dataValues.currency
            isClientCurrencyBtc = 
            paymentCurrency === `BTC` ||
            paymentCurrency === `SATS`
        }

    
```

api/models/Payment.js
```diff
@@ -17,6 +17,10 @@ module.exports = (sequelize) => {
             type: DataTypes.INTEGER,
             allowNull: false
         },
+        currency: {
+            type: DataTypes.STRING,
+            allowNull: false
```
otech47, line 22: For eventual production migration, this will fail because existing payments will not have a currency

We should either allow null values or have a separate migration that looks at every payment, finds the currency through the Client and apply it to the payments column

PR #788, opened 2023-04-21
Running both UI's

sofiaromorales:


Ovifer13 (commented):
I think we also need to change `SITE_ROOT` in `constants.js`, and do something to validate that we are using 6003 or 6002.
I was researching to see if there's a way to know what port you are in and change the conditions, instead of 6002, redirect to 6003.
https://github.com/setlife-network/trinary/blob/1b698162d6632ff783d36850310df920808ced3e/api/config/constants.js#L6-L8

sofiaromorales:
> I think we also need to change `SITE_ROOT` in `constants.js`, and do something to validate that we are using 6003 or 6002. I was researching to see if there's a way to know what port you are in and change the conditions, instead of 6002, redirect to 6003.
> 
> https://github.com/setlife-network/trinary/blob/1b698162d6632ff783d36850310df920808ced3e/api/config/constants.js#L6-L8

After discussing this with @otech47 we decided to leave this as it is, for now, the problem with the redirect only happens when logging in, so not a big deal

PR #789, opened 2023-04-24
V1 working with client_id

### LINKED ISSUE #787: Project without client id
In UIv1 you can only create a project within a client
In UIv2 is possible to create projects without a client id assigned to it, generating breaks in UIv1

This task is about refactoring UIv1 so it doesn't break when trying to access a project without a client.

Ovifer13:
**Issue #787**

- [x] Fix errors caused by `client_id`
- [x] Fix all other errors related with client-project relation

### DIFF
--- a/api/schema/resolvers/ProjectResolver.js
+++ b/api/schema/resolvers/ProjectResolver.js
@@ -564,15 +564,22 @@ module.exports = {
             return models.Project.findAll()
         },
         getActiveProjects: (root, args, { models }) => {
+
             return models.Project.findAll({
                 where: {
-                    is_active: true
+                    is_active: true,
+                    client_id: {
+                        [Op.is]: !null
+                    }
                 }
             })
         },
         getActiveProjectsCount: (root, args, { models }) => {
             const whereFields = {
-                is_active: true
+                is_active: true,
+                client_id: {
+                    [Op.is]: !null
+                }
             }
             if (args.clientId) whereFields.client_id = args.clientId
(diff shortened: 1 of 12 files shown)


otech47:
@Ovifer13 you can also consider in this refactor instead of finding all the places in the UI that might break if `project.client` is NULL, you can just tweak all of the fetch project queries to make sure they only return projects with a NON-NULL client_id.

whichever way is more convenient for you is fine

Ovifer13:
@otech47 that solution looks like it's working but we won't be able to see projects without clients, in the future we could but we will have to refactor a lot of coding for what I was facing, there are a lot of components that depend on `client`.

otech47:
> @otech47 that solution looks like it's working but we won't be able to see projects without clients, in the future we could but we will have to refactor a lot of coding for what I was facing, there are a lot of components that depend on `client`.

Yeah that's fine for v1

api/schema/resolvers/ProjectResolver.js
```diff
@@ -564,15 +564,22 @@ module.exports = {
...
+                    }
                 }
             })
         },
         getActiveProjectsCount: (root, args, { models }) => {
             const whereFields = {
-                is_active: true
+                is_active: true,
+                client_id: {
+                    [Op.is]: !null
```
sofiaromorales, line 580: Does this filter only apply to UI V1 @Ovifer13?

sofiaromorales (approved):

PR #801, opened 2023-05-11
Wallet API

### LINKED ISSUE #799: Create `Wallet` type and resolver

sofiaromorales:
### **Issue #799**
Screen Shot 2023-05-11 at 2 46 21 PM
Screen Shot 2023-05-11 at 2 46 53 PM

@Ovifer13 I believe this implementation is enough for setting up the wallet, let me know what you think!


Ovifer13 (approved):
It should be good like this! Just want to know, when does the user get the `onchain_address`?

sofiaromorales:
> It should be good like this! Just want to know, when does the user get the `onchain_address`?

In the simple option, on-chain address is basically copying and pasting the address from your wallet

PR #824, opened 2023-06-01
API mutation to transform USD to SATS

### LINKED ISSUE #821: Build API query to transform USD to SATS
When sending bonuses the amount will be displayed in USD, but the confirmation will be in SATS, so it's necessary to build a mutation that given a USD amount returns the SATS amount

Ovifer13:
**Issue #821**

- [x] Create mutation to exchange from USD to SATS

**Proof**
image


sofiaromorales (approved):

sofiaromorales:
@ovifer why using `mutation` instead of `query`?

